PR #115, opened 2015-03-05
Create new ButtonEditLink component to use in case of link selection and link edition.

jbalsas:
Hey Iliyan,

This is a possible implementation for the Link Button. I agreed some of the behaviour changes with Juan. 

I've added the react+addons bundle so we can easily use `React.addons.LinkedStateMixin` among other sugar stuff ;) If you prefer, we can implement it on our own using listeners and state.

Let me know what you think


jbalsas:
I've just pushed one extra commit with the Twitter Button as well!

src/ui/react/src/components/buttons/button-twitter.jsx
```diff
@@ -0,0 +1,92 @@
...
+         * Generates the appropriate twitter url based on the selected text and the configuration
+         * options received via props.
+         *
+         * @return {String} A valid twitter url with the selected text and given configuration.
+         */
+        _getHref: function() {
+            var nativeEditor = this.props.editor.get('nativeEditor'),
+                selectedText = nativeEditor.getSelection().getSelectedText(),
+                url = this.props.url,
+                via = this.props.via,
```
jbalsas: We need to get back the ability to specify buttons as objects rather than just strings so we can pass additional configuration values to them.

ipeychev:

PR #291, opened 2015-07-21
Allow to prevent AlloyEditor from setting contenteditable to true on its srcNode

### LINKED ISSUE #290: Keep the contenteditable attribute value
In some cases, you don't want the node passed to `AlloyEditor.editable` to be completely editable. Basically instead of forcing `contenteditable` to true, AlloyEditor should only set it if it's not already set. (PR to come shortly)

dpobel:
Fixes #290 

Note: testing that requires quite some changes in the alloy-editor tests but I'm unsure if the test structure in this patch is the best one.

### DIFF
--- a/src/ui/react/src/adapter/alloy-editor.js
+++ b/src/ui/react/src/adapter/alloy-editor.js
@@ -23,7 +23,10 @@
          */
         initializer: function(config) {
             var node = this.get('srcNode');
-            node.setAttribute('contenteditable', 'true');
+
+            if ( this.get('enableContentEditable') ) {
+                node.setAttribute('contenteditable', 'true');
+            }
 
             var editor = CKEDITOR.inline(node);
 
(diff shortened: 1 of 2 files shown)


ipeychev:

dpobel:
I updated the patch after the discussion in #290 but this does not work yet because of a bug in the implementation of `writeOnce` attribute together with a default value. Because of the order of the conditions https://github.com/liferay/alloy-editor/blob/master/src/ui/react/src/oop/attribute.js#L140, the default value always overrides the value provided in the configuration.

How should I proceed with this bug affecting the current patch ? Should I open a separate issue for this one ? Can I fix this issue in this PR or should I open a different one ? And last, are there any tests on that ?

ipeychev:
Hey @dpobel 

Thanks for the patch! In cases like this, we open a new issue, fix it, create a test and then we continue with the original one. In short - you will have to create another issue.

And yes, we have multiple [tests](https://github.com/liferay/alloy-editor/blob/master/src/ui/react/test/attribute.js) for the Attribute, so I am a bit surprised that we have a bug there, but life if is like this.

dpobel:
Actually, in the corresponding test, the assertion reproduces the bug https://github.com/liferay/alloy-editor/blob/master/src/ui/react/test/attribute.js#L267. Anyway, I'll fix that in different issue and PR.

dpobel:
Rebased against master to get the fix for the writeOnce attribute default value and the tests are all green on my setup at least.

ipeychev:

ipeychev:
Thank you, pull request merged! [See changes here.](https://github.com/liferay/alloy-editor/compare/7834b51~3...7834b51) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

PR #303, opened 2015-07-27
Skip the editorInteraction event when there is no native selection.

### LINKED ISSUE #302: Uncaught TypeError: Cannot read property 'createRange' of null
I get this error at during various user interactions. It's hard to pinpoint the type of or order of events that triggers it.

It's not always `createRange`, but it's always a range related name.

When these errors happen, nothing in the UI or editable block appears to break. So this seems to be a case of something being triggered when it isn't needed and not validating things before using it. Which probably can be fixed simply by testing before doing things.

((Note that personally that kind of error is a bigger problem. Because I treat uncaught errors as a reason to warn the user of a potential bug and recommend a refresh.))

```
Uncaught TypeError: Cannot read property 'createRange' of null
  window.AlloyEditor.CKEDITOR.plugins.get.e.getClientRectsRegion @ VM4516:680
  window.AlloyEditor.CKEDITOR.plugins.get.e.getSelectionRegion @ VM4516:680
  window.AlloyEditor.CKEDITOR.plugins.get.e.getSelectionData @ VM4516:680
  (anonymous function) @ VM4516:680
  (anonymous function) @ VM4516:680
  setTimeout (async)
  window.AlloyEditor.CKEDITOR.tools.debounce.CKEDITOR.tools.debounce.r @ VM4516:680
  j @ VM4516:10
  CKEDITOR.event.CKEDITOR.event.fire @ VM4516:12
  CKEDITOR.dom.domObject.a @ VM4516:45
```

dantman:
This fixes #302.

### DIFF
--- a/src/core/selection-region.js
+++ b/src/core/selection-region.js
@@ -144,6 +144,10 @@
         getSelectionData: function() {
             var selection = this.getSelection();
 
+            if ( !selection.getNative() ) {
+                return;
+            }
+
             var result = {
                 element: selection.getSelectedElement(),
                 text: selection.getSelectedText()
--- a/src/core/uicore.js
+++ b/src/core/uicore.js
@@ -83,10 +83,13 @@
                         ariaState = [];
 
                         if (event.name !== 'keyup' || event.data.$.keyCode !== 27 || editor.config.allowEsc) {
-                            editor.fire('editorInteraction', {
-                                nativeEvent: event.data.$,
-                                selectionData: editor.getSelectionData()
-                            });
+                            var selectionData = editor.getSelectionData();
+                            if ( selectionData ) {
+                                editor.fire('editorInteraction', {
+                                    nativeEvent: event.data.$,
+                                    selectionData: selectionData
+                                });
+                            }
                         }
                     },
                     uiTasksTimeout


ipeychev:

ipeychev:
There are other public functions, which depend on the native selection, so we should fix them too.
For that purpose, let's go with the following [commit](https://github.com/liferay/alloy-editor/commit/ba45a6f5cfc865c6297ad7f30acde03a145d1cbc).

Thank you very much for the contribution!

ipeychev:

ipeychev:
Hey @dantman,

After the changes I [committed](https://github.com/liferay/alloy-editor/commit/ba45a6f5cfc865c6297ad7f30acde03a145d1cbc), the exception does not appear anymore, but your PR actually could be useful in order to resolve an UX issue, so I will merge it. Basically, at least in Chrome and using the current master, if you make a selection (so the toolbars appear) and switches to another application using Cmd/Ctrl + Tab, the toolbars will appear outside of the editor. This is due to the default values we set to the selection data object.

Some notes about the commits we make in AlloyEditor:
1) In general, we have a strict rule - a commit and especially bugfix should have tests. An exception of this rule is when we just cannot test it. I think this issue falls exactly in this category. I tried several ways, but I wasn't able to force all browsers to loose the native selection. If you can think of a way to test it, that would help a lot!

2) All commit messages should include something like:
```
Fixes #303
```

so we will know what was the purpose of the commit and to which ticket it belongs.
I amended your commit message, this was just FYI.

Thanks again for your contribution, we really appreciate it!

Iliyan

ipeychev:
Thank you, pull request merged! [See changes here.](https://github.com/liferay/alloy-editor/compare/3ce8667~2...3ce8667) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

PR #352, opened 2015-10-06
Moved the contenteditable test to the selection test method

dpobel:
A check against the `isContentEditable` property of the target element was added in the Styles toolbar render method to avoid letting the user style a non editable element. In most cases that's perfectly fine but if you use the widget CKEditor plugin this becomes a problem because this plugin wraps the widget HTML element in a non editable element while you might still want to be able to style the widget with some (custom) buttons.

~~To workaround this problem, this patch adds the support for an optional `canStyleElement` config property in the Styles toolbar so that it's possible to pass a custom function to decide whether the element can be styled or not.~~

~~This does not break any tests and if the approach is OK, I'll of course add some unit tests for that feature.~~

The new version of this patch moves the content editable check to the selection test methods.

### DIFF
--- a/src/core/table.js
+++ b/src/core/table.js
@@ -1,6 +1,18 @@
 (function() {
     'use strict';
 
+    var IE_BUGGY_EDITABLE_ELEMENT = {
+        "table": 1,
+        "col": 1,
+        "colgroup": 1,
+        "tbody": 1,
+        "td": 1,
+        "tfoot": 1,
+        "th": 1,
+        "thead": 1,
+        "tr": 1
+    };
+
     /**
      * Table class utility. Provides methods for create, delete and update tables.
      *
@@ -92,6 +104,29 @@
             }
 
             return table;
+        },
+
+        /**
+         * Checks if the given table can be considered as editable. This method
+         * method is there to workaround a limitation of IE where for some
+         * elements (like table), `isContentEditable` is always false because IE
+         * does not support contenteditable on such elements but that does not
+         * really mean the content of such elements can not be edited.
+         *
+         * See https://msdn.microsoft.com/en-us/library/ms537837%28v=VS.85%29.aspx
+         *
+         * @method isEditable
+         * @param {CKEDITOR.dom.element} el the table element to test
+         * @return {Boolean}
+         */
+        isEditable: function (el) {
+            if ( !CKEDITOR.env.ie || !el.is(IE_BUGGY_EDITABLE_ELEMENT) ) {
+                return !el.isReadOnly();
+            }
+            if ( el.hasAttribute('contenteditable') ) {
+                return (el.getAttribute('contenteditable') !== 'false');
+            }
+            return this.isEditable(el.getParent());
         },
 
         /**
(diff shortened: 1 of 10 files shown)


dpobel:
Hi @ipeychev (glad you're still around :))

I already use several custom `test` functions to handle various types of selection but in the case where the user gives the focus to a widget the test function is not even called because of the condition on `isContentEditable` in the very beginning of the `render` method. 
That said, I see your point and I kind of agree. So maybe a better solution would be to move the `isContentEditable` check in the existing test functions and let the custom ones decide whether the content editable property really matters ?

ipeychev:
Hey @dpobel

Yes, I'm around. I left Liferay, but AlloyEditor is in my heart so I'm going to contribute to it as much as possible. Moreover, I'm still the maintainer.

So, about the issue, I think this makes sense. Let's move it inside the test functions. Are you willing to make this change?

Thanks,

dpobel:
yes, I'll try to work on that today, thanks for your feedback.

dpobel:
I updated the patch. While I'm working on the unit tests, I would be happy to get some feedback.

ipeychev:

dpobel:
I finished implementing the unit tests also did the optimizations suggested by @ipeychev. I guess this is ready for a final review and hopefully a merge :)

ipeychev:
Great, @dpobel, thanks!
Just to note: we will be able to merge this PR probably in a few days - we are in process of updating the permissions to the repo. Nothing scary.

andrerom:
:+1: 
@ipeychev When merged, will there be possibility to get this as part of a tag some time soon as well?

ipeychev:
Yeah, version 0.6 will be released after that!

ipeychev:

ipeychev:
@dpobel 

Great work, thanks! There is a test, which fails on IE (9, 10 and 11). It is [this](https://github.com/ezsystems/alloy-editor/commit/95c86e8f656e909ff2c8999be96b277132e143bd#diff-f962a7400f899491a2fe2c09979386a5R180) one. Does it pass for you?

Thanks,

dpobel:
:( I'm working on Linux so I did not run the test in IE (my mistake). I'm gonna setup a VM and try to see what's going on.

dpobel:
https://msdn.microsoft.com/en-us/library/ms537837%28v=VS.85%29.aspx
> Though the TABLE, COL, COLGROUP, TBODY, TD, TFOOT, TH, THEAD, and TR elements cannot be set as content editable directly [...]

so I have to workaround that somehow

ipeychev:
Okay, great, check if you can figure it out, I don't have a suggestion on top of my head.

dpobel:
So I added a `isEditable` method in `CKEDITOR.Table` and use in the test selection for table. But this was only tested in IE11 but hopefully it works in IE9 and IE10.

ipeychev:

ipeychev:
Thank you, pull request merged! [See changes here.](https://github.com/liferay/alloy-editor/compare/4504257~2...4504257) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

ipeychev:
Great job @dpobel!
:+1:

PR #354, opened 2015-10-09
Convert buffers toString before templatizing

gerad:
**why**

Recent versions of node read files as buffers (arrays of bytes) the template
library was treating these as arrays when inlining them, resulting in invalid
output.

**before**

**after**


ipeychev:

ipeychev:
Thank you, pull request merged! [See changes here.](https://github.com/liferay/alloy-editor/compare/636d1be~2...636d1be) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

ipeychev:
Hey @gerad,

Thanks for your contribution! I modified a bit the commit message, because:
1. We always mention the issue, to which this commit is associated. In this case we can track comments, or, like in your case, to publish screenshots, etc.
2. We don't publish links inside commit messages. A commit message should contain only plain text.

I also dropped the commit with build files and added a new one. Only the maintainer can publish build files. This is to minimize the risk of adding malicious code to the repo.

Thanks again for your contribution, we really appreciate it!

Iliyan

gerad:
:+1: will confirm for next time

PR #368, opened 2015-11-02
Fire editorUpdate event when UI state is updated

jbalsas:
Hey @ipeychev, what do you think about this? Do you think firing on `componentDidUpdate` could cause some performance impact? Should we have an attribute to control wether we fire the event or not?

In addition... I think it's time we get some event emitter in there so we can start firing events from the AlloyEditor instance instead of CKEditor. What do you think?

### DIFF
--- a/src/ui/react/src/components/main.jsx
+++ b/src/ui/react/src/components/main.jsx
@@ -117,11 +117,20 @@
         componentDidUpdate: function (prevProps, prevState) {
             var domNode = React.findDOMNode(this);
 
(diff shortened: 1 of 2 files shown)


ipeychev:

ipeychev:
Thank you, pull request merged! [See changes here.](https://github.com/liferay/alloy-editor/compare/452e69e~2...452e69e) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

PR #369, opened 2015-11-02
Make ButtonStyle.isActive() work if there is no selection.

arendjr:
This makes it possible to create style buttons in a toolbar that is always visible, rather than the default toolbars that only appear when you select something.

### DIFF
--- a/src/ui/react/src/components/base/button-style.js
+++ b/src/ui/react/src/components/base/button-style.js
@@ -58,6 +58,9 @@
             var editor = this.props.editor.get('nativeEditor');
 
             var elementPath = editor.elementPath();
+            if (!elementPath) {
+                return false;
+            }
 
             result = this.getStyle().checkActive(elementPath, editor);
 
--- a/src/ui/react/src/components/buttons/button-styles.jsx
+++ b/src/ui/react/src/components/buttons/button-styles.jsx
@@ -121,13 +121,18 @@
         _checkActive: function(styleConfig) {
             var nativeEditor = this.props.editor.get('nativeEditor');
 
+            var elementPath = nativeEditor.elementPath();
+            if (!elementPath) {
+                return false;
+            }
+
             // Styles with wildcard element (*) won't be considered active by CKEditor. Defaulting
             // to a 'span' element works for most of those cases with no defined element.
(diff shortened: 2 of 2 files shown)


ipeychev:

ipeychev:
Hey @arendjr,

Thanks a lot for your contribution, we highly appreciate it!
Please let me share the rules we all follow when we make contributions to this project:
1. The first rule and the most important one is that every change should have a purpose and this purpose should be clear for everyone.
The written explanation is:
"_This makes it possible to create style buttons in a toolbar that is always visible, rather than the default toolbars that only appear when you select something._"
I would appreciate some addition details and tests, because right now for me there is no way to reproduce the issue, hence, to spot if that change breaks something else. The commit message is:
"_Make ButtonStyle.isActive() work if there is no selection._". However, what was the issue previously? What means that "ButtonStyle.isActive()" does not work? Does it throw an exception, or it just does not return proper value?

Could you please clarify?

Thanks,
Iliyan

arendjr:
Hi there,
Sorry to inform you, but I've stopped my evaluation of AlloyEditor. I was attempting to use it with a more traditional toolbar above the editor, like you see above many other editors. I disabled the floating toolbars and created my own using the React components you provide. When doing this I quickly ran into the issue this PR is meant to solve: If you render any of the buttons while the cursor is not currently in the editor, you will get an exception thrown. This occured because `editor.elementPath()` returns `null` (or maybe it was `undefined`) in such cases which would cause an error inside the `checkActive()` method.
Unfortunately, once these issues were fixed I ran into other, more obscure exceptions which led me to believe the AlloyEditor is just not intended to be used the way I was trying to and the effort it would take to get it to work reliably in a way that's so much different from how you guys intent to use it might not be worth it.
Hope you might still see some value in the PR if you're interested in the use case for traditional toolbars, but I can understand if you wish to ignore it otherwise :)
Cheers,
Arend jr.

ipeychev:
Okay, thanks for your time and explanation!
AlloyEditor is indeed not meant to be used with fixed toolbar. We don't believe in this concept and we don't think it makes sense. We think the present and the future belongs to editors which provide context depending toolbar and widgets.

Thanks again for your time!

PR #370, opened 2015-11-04
Update tests to fix breaking changes and remove deprecation warnings

jbalsas:
Hey @ipeychev, 

This PR updates our React version to current `0.14.2`

Basically, what I've done is:
- Grab new versions and add a build step to concat the extracted `react-dom` module
- Update source code to remove all deprecation warnings so we'll be ready when they break
- Update tests to fix the introduced breaking changes and remove the deprecation warnings

In the end, it's not such a big of a change, but I'd still recommend we go for a major version change whenever we merge this since it will break for sure for people using `alloyeditor-no-react` and their own `React` version.

What do you think?


ipeychev:

ipeychev:
Hey @jbalsas,

Great, this was really cool! I added some additional changes:
1. Change the UMD header in React and ReactDOM - otherwise, we wouldn't be able to require alloy-editor-all in NodeJS. We have to do this when we upgrade React.
2. Update the build task in order to properly create minimized version of React.

Yeah, it seems version 0.7 will be needed, because the changes in React are too much and we also had to make changes. There is now an additional property - AlloyEditor.ReactDOM, so we definitely go to 0.7.

Thanks!

jbalsas:
Hey Iliyan, good catch! I completely missed those! :disappointed: 

Maybe we could design some tests to at least simulate our supported environments so I won't forget that next time

ipeychev:
Yeah, what we need is actually something very simple:
```
require('../alloy-editor-all')
```
This shouldn't throw any exception. However, this test should run on the server, not in Karma

Issue #371, opened 2015-11-04
toolbar does not disappear


kel093:
Hello. I am apologise if i duplicate the issue, but i don't found it in list of issue's. 
I am using the AlloyEditor and today saw strange behaviour. I selected the text after this the toolbar is appears, then i clicked on the any place on the screen and the toolbar does not disappear.
![bug](https://cloud.githubusercontent.com/assets/6705879/10938785/733cd998-8305-11e5-8f06-c7c6d1b3e051.png)

I investigated how plugin works and understood the problem, the problem is that editor.status == "ready" and listener "contentDom" will not be called. That's means that listeners "mousedown" and "keydown" are not be added to document.

I fixed this problem, please see my changes in function componentDidMount().

            if (editor && editor.status == "ready") {
                document.addEventListener('mousedown', this._mousedownListener);
                document.addEventListener('keydown', this._keyDownListener);
            } else {
                editor.once('contentDom', (function () {
                    document.addEventListener('mousedown', this._mousedownListener);
                    document.addEventListener('keydown', this._keyDownListener);
                }).bind(this));
            }

ipeychev:
Hey @kel093,

Interesting issue! How exactly we can reproduce it?
Also, could you see if the following code snippet will fix your issue too:
```
editor.once('instanceReady', function() {
    document.addEventListener('mousedown', this._mousedownListener);
    document.addEventListener('keydown', this._keyDownListener);
}.bind(this));
```

Maybe we can just change the event and avoid the other "if".

Thanks,

kel093:
I tried to change event "contentDom" on "instanceReady", but didn't help.

It's easy to reproduce, just call destroy() then init() on earlier initialized AlloyEditor.

Look like on this code:
<pre lang="javascript"><code>
var editor2 = AlloyEditor.editable('editable'); //init the AlloyEditor
setTimeout(function(){
    //reinit AlloyEditor
    editor2.destroy();
    editor2.init();
},1000);
</code></pre>

ipeychev:
Ah, I see! This way of reinitializing is unsupported however. The function "init" is not supposed to be called manually.
The recommended way after destroying is to get a new fresh instance of the editor, like this:
```
var editor2 = AlloyEditor.editable('editable'); //init the AlloyEditor
setTimeout(function(){
    //reinit AlloyEditor
    editor2.destroy();
    editor2 = AlloyEditor.editable('editable');
},1000)
```

Thanks,

kel093:
[...]

PR #373, opened 2015-11-05
Use focus event instead of contentDOM to make sure initialization always runs

jbalsas:
Hey @ipeychev, this is the fix for #371.

I've been fighting to get a proper test for a while and I can't spend much more time on it. On one side, `React.TestUtils` can't simulate pure DOM events, so we have to use CKEditor jquery or add something else.

Also, for some reason, I was observing a weird initialization order when destroying and recreating an AlloyEditor instance, which prevented me from covering the use case where we destroy and recreate the editor.


ipeychev:

PR #380, opened 2015-11-19
Do not prevent plugin from being created if original CKEditor one exists

### LINKED ISSUE #379: Plugins 'ae_' are not initialized if ckeditor own plugins are present
When using the `alloyeditor-no-ckeditor` version, some plugins might conflict with the existing one. This affects

- `dragresize`
- `tableresize`
- `tabletools`

If the CKEditor installation has the original bundled plugins, the namespaced ones won't get created. Then, if the editor is instantiated doing the following (not code complete):

```javascript
    extraPlugins: 'ae_dragresize,ae_tableresize,ae_tabletools',
    removePlugins: 'dragresize,tableresize,tabletools'
});
```

The namespaced plugins won't produce any behaviour since we are returning empty objects for them.

Since we namespaced them, it should be safe enough to avoid the check for the original ones.

jbalsas:
Hey @ipeychev this fixes #379.

I tested this using a custom build of CKEditor with all of the default plugins. To be able to create proper tests, we'd have to commit that one and use it as part of the test process, so we'd have to keep two copies of CKEditor in our repo. 

Do you think it is worth keeping it for this, or future testing purposes?

### DIFF
--- a/src/plugins/dragresize.js
+++ b/src/plugins/dragresize.js
@@ -9,7 +9,7 @@
 (function() {
     'use strict';
 
-    if (CKEDITOR.plugins.get('dragresize') || CKEDITOR.plugins.get('ae_dragresize')) {
+    if (CKEDITOR.plugins.get('ae_dragresize')) {
         return;
     }
 
--- a/src/plugins/tableresize.js
+++ b/src/plugins/tableresize.js
@@ -6,7 +6,7 @@
 (function() {
     'use strict';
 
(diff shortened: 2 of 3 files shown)


ipeychev:

ipeychev:
Thank you, pull request merged! [See changes here.](https://github.com/liferay/alloy-editor/compare/aaf1b97~1...aaf1b97) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

Issue #381, opened 2015-11-20
ERR_INVALID_URL when moving an image in the content


projer:
Hi!

First I wanted to congratulate you for your RTE, very nice job!

I was testing it, and I found that on Chrome and Opera when you paste an image inside and move it, after the second time, an ERR_INVALID_URL appears in the console, which results in inserting strange characters (â€‹), and even from time to time it duplicates the image.

Looks like it works perfectly on Firefox.

Thanks!
![2015-11-20_15h31_41](https://cloud.githubusercontent.com/assets/3242718/11302137/ddb6354c-8f9b-11e5-8745-06c680c4245c.png)

ipeychev:
Hey @projer,

Thanks for the reporting and for the nice words!

This issue seems to be not in our code but in the CKditor's engine, which we use. When that happens, here is how we proceed:
- If the bug is critical, we report it and meanwhile we try to fix it by ourselves or at least to narrow it down.
- If not critical, we report it and upgrade to a new version when they fix it.

Since the normal behavior is the image to be immediately uploaded and with a "real" URL the bug cannot be reproduced, I will consider this bug as not critical.

Thanks,

ipeychev:
Chema, could you please go ahead and update CKEditor to their latest version (currently 4.4.5)? It seems this issue is fixed there and the upgrade will fix #375 too (they now throw a warning - http://docs.ckeditor.com/#!/guide/dev_errors-section-editor-incorrect-destroy)

@jbalsas 
Thanks,

jbalsas:
Hey @ipeychev,

I've updated the version [here](https://github.com/liferay/alloy-editor/pull/382)

I tried it out and indeed the error stops showing, but because the behaviour with CKEditor 4.5.5 seems to be that images with base64 content as urls can not be moved around.

PR #382, opened 2015-11-21
Update lang files

### LINKED ISSUE #375: Bug: calling editor.destroy() throws javascript error
Hey guys,

Found a small issue. To reproduce, try this: https://github.com/brunobasto/alloy-editor/commit/9b71aff126d1f755ae90b107f8b8161c165927d8

Thanks,

jbalsas:
Hey @ipeychev, I've updated CKEditor to 4.5.5. This should fix #375 and #381.

I'll try to keep our version in sync with CKEditor releases moving forward to avoid having unfixed issues in our branch.

Thanks!


src/ui/react/src/adapter/alloy-editor.js
```diff
@@ -99,22 +101,24 @@
          * @method _renderUI
          */
         _renderUI: function() {
-            var editorUIElement = document.createElement('div');
-            editorUIElement.className = 'ae-ui';
+            if (!this._destroyed) {
```
jbalsas: @ipeychev, can you think of a better way of doing this?

The thing is that because we call `_renderUI` asynchronously on `langResourcesLoaded`, it might still happen that we render the UI after the `destroy` method has been called.

The visible result is that, even if CKEditor does not complain now, we still get issues because of the `_mousedown` events we attached and couldn't remove because of `destructor` being invoked first.

ipeychev:

ipeychev:
Hey Chema,

I can imagine, but not with the current logic/code. We don't have mechanism for canceling the process of loading resources, so on practice there is nothing to do. We could introduce such, of course, but I don't think it is worth the efforts. The solution you suggest is the cheapest one, so let's go with it.

Thanks,

ipeychev:
Thank you, pull request merged! <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

PR #392, opened 2015-12-10
Update Language files (auto-generated)

jbalsas:
Hey @ipeychev 

This PR does several things:
- Updates the link util to support working with link attributes
- Changes the link selection test to work on empty selections (on click)
- Adds a dropdown inside the link edit button to change the link target attribute (enabled by default)

I think this solves some of the UX issues regarding links we've been discussing.

I wasn't sure about the defaults. For example, we could choose to hide the target button by default instead, so we keep the UI simpler. We could also keep the link selection test as it is but provide the click one as part of the bundle so one could choose to change the behaviour...

What do you think?

Thanks!


ipeychev:

ipeychev:
Thank you, pull request merged! [See changes here.](https://github.com/liferay/alloy-editor/compare/3494b6c~2...3494b6c) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

ipeychev:
Hey @jbalsas,

This is really cool! About hiding the target by default, the UI would really become simpler, but I think is not that complex even now. Let's see what the community will say, I am OK with that UI as a user. The same is valid for the link selection test, let's do that if needed.

Thanks!

jbalsas:
Glad you like it! :)

@ipeychev, how do you feel about releasing a new version? I think we've added several things already that grant it. I'm not sure if it should be a minor or a major version, though. What do you think?

ipeychev:
I was going to release a minor version, Chema, but the build failed :(
https://travis-ci.org/liferay/alloy-editor/builds/96473449
There is one test, which fails on IE11 on Windows 8.1. On my machine, where I have a Windows 7 running via VirtualBox, they pass. i'm currently wondering what to do.

jbalsas:
Hey Iliyan, I'm looking at it, and it actually makes sense that it could fail. What it's actually weird is that it works for the others.

The `linkSelectionTest` never uses the payload to test the element, but grabs it from the editor itself. The same could potentially happen to `textSelectionTest` and `tableSelectionTest`

https://github.com/liferay/alloy-editor/blob/master/src/ui/react/src/selections/selection-test.js#L11
https://github.com/liferay/alloy-editor/blob/master/src/ui/react/src/selections/selection-test.js#L37
https://github.com/liferay/alloy-editor/blob/master/src/ui/react/src/selections/selection-test.js#L47

Not sure if we could do a better job of detecting the element in https://github.com/liferay/alloy-editor/blob/master/src/core/selection-region.js#L163-L166

jbalsas:
I think something like this should make things more consistent

https://github.com/jbalsas/alloy-editor/commit/573020911f11230013c4edc1d4cfeb38676bd738

If you agree I can apply it to the other entities we have and send a PR.

ipeychev:
Ah, it fails on Firefox too! On my machine I just updated the packages and it revealed it.

CKEditor's `selection.getSelectedElement()` on theory should do the trick... maybe they have some bug, which was already reported?
As fallback for now we can use our Link, it won't hurt at all. Did you actually try this code?

jbalsas:
Yeah, I tried it and it seemed to be working. 

I don't think CKEditor's `selection.getSelectedElement()` has ever done that. It only returns elements such as images, tables in theory, hr (surprisingly this does work :D) ...

ipeychev:
Hmm, but I think `getSelectedElement` does checks for link? See their code here:
http://docs.ckeditor.com/source/selection.html#CKEDITOR-dom-selection-method-getSelectedElement
There is a reference to `styleObjectElements `, which is:
```javascript
var styleObjectElements = { img: 1, hr: 1, li: 1, table: 1, tr: 1, td: 1, th: 1, embed: 1, object: 1, ol: 1, ul: 1,
    a: 1, input: 1, form: 1, select: 1, textarea: 1, button: 1, fieldset: 1, thead: 1, tfoot: 1 };
```

There is `a` element in this object?

In any case, I'm okay if we can do it by ourselves. What other stuff it breaks?

jbalsas:
Mmmm... yeah, I think it can do it, but it's tricky to get the right selection manually. It's easy with `image` and `embed`, but very hard for the rest of elements.

It kind of breaks the `textSelectionTest` that expects currently that `!selectionData.element`. With this, selecting a piece of a link will fail the `selectionIsEmpty` check of the link plus that one of the text :cry: 

Maybe what we need to change is the test itself... since the selection test functions do not only depend on the payload, we should take that into account to design those tests as well...

jbalsas:
Hey @ipeychev this fixes the test in Firefox for me https://github.com/jbalsas/alloy-editor/commit/dbe79e5bde12f339e43daada4c52746bf064b363

It also popped up here when I updated the dependencies... 

I don't have the IE vms in this laptop, so I can't test that it works for IE, but I think it should...

ipeychev:
Hey Chema,

It worked, great! Will commit this and set you as the author of the commit ;)
I will do an additional commit with updating the dependencies, because additional changes should be done.

Issue #128, opened 2015-03-12
React UI Edit Link should move cursor after link creation


jbalsas:
In the ReactUI, after creating a link, the selection is retained and the link button shows. It would be better if the selection was moved after the created/updated link so the user can keep writing.

jbalsas:
Hey @ipeychev,

Now that we've changed the link selection, it could make sense to rethink this. This is the current state:

- After creating a new link, we change the selection to get all the element. This now causes the text selection buttons to show in the styles toolbar. (this looks good to me)
- After updating an existing link, we keep the empty selection, so the edit UI doesn't go away. 

The way I see it, we could either select the link on update to match the create behaviour, or advance the cursor to the next position after the link as suggested in #387 to dismiss the styles toolbar.

What do you think?

ipeychev:
Hey Chema,

I would expect the toolbar to disappear in both cases. After all, the primary operation - creation/editing the link has been completed. After creation, it is quite unusual to modify the style, so despite it is convenient that the toolbar stays there, I feel that most of the users won't touch it, so probably this will only force them to click one more time.
On editing, is the same logic. The goal - link editing has been completed, so I would expect the toolbar to disappear.

Thanks,

jbalsas:
Hey Iliyan,

Makes sense, it's also aligned with what was being asked in #387.

Will work on that direction then :blush:

PR #397, opened 2015-12-13
Advance cursor after link on link changes

jbalsas:
Hey @ipeychev, this could be a fix for #128 

I'm going out now, so I'm just sending this to get your opinion. I'll add tests and complete the PR tomorrow from the office :wink:

### DIFF
--- a/src/core/link.js
+++ b/src/core/link.js
@@ -19,13 +19,44 @@
         constructor: Link,
 
         /**
+         * Advances the editor selection to the next available position after a
+         * given link or the one in the current selection.
+         *
+         * @param {CKEDITOR.dom.element} link The link element which link style should be removed.
+         * @method advanceSelection
+         */
+        advanceSelection: function(link) {
+            link = link || this.getFromSelection();
+
+            var range = this._editor.getSelection().getRanges()[0];
+
+            if (link) {
+                range.moveToElementEditEnd(link);
+
+                var nextNode = range.getNextEditableNode();
+
+                if (nextNode && !this._editor.element.equals(nextNode.getCommonAncestor(link))) {
+                    var whitespace = /\s/.exec(nextNode.getText());
+
+                    var offset = whitespace ? whitespace.index + 1 : 0;
+
+                    range.setStart(nextNode, offset);
+                    range.setEnd(nextNode, offset);
+                }
+            }
+
+            this._editor.getSelection().selectRanges([range]);
+        },
+
+        /**
          * Create a link with given URI as href.
          *
          * @method create
          * @param {String} URI The URI of the link.
          * @param {Object} attrs A config object with link attributes. These might be arbitrary DOM attributes.
+         * @param {Object} modifySelection A config object with an advance attribute to indicate if the selection should be moved after the link creation.
          */
-        create: function(URI, attrs) {
+        create: function(URI, attrs, modifySelection) {
             var selection = this._editor.getSelection();
 
             var range = selection.getRanges()[0];
@@ -50,7 +81,12 @@
 
             style.type = CKEDITOR.STYLE_INLINE;
             style.applyToRange(range, this._editor);
-            range.select();
+
+            if (modifySelection && modifySelection.advance) {
+                this.advanceSelection();
+            } else {
+                range.select();
+            }
         },
 
         /**
@@ -83,12 +119,17 @@
          * Removes a link from the editor.
          *
          * @param {CKEDITOR.dom.element} link The link element which link style should be removed.
+         * @param {Object} modifySelection A config object with an advance attribute to indicate if the selection should be moved after the link creation.
          * @method remove
          */
-        remove: function(link) {
+        remove: function(link, modifySelection) {
             var editor = this._editor;
 
             if (link) {
+                if (modifySelection && modifySelection.advance) {
+                    this.advanceSelection();
+                }
+
                 link.remove(editor);
             } else {
                 var style = new CKEDITOR.style({
@@ -114,8 +155,9 @@
          * @method update
          * @param {Object|String} attrs The attributes to update or remove. Attributes with null values will be removed.
          * @param {CKEDITOR.dom.element} link The link element which href should be removed.
+         * @param {Object} modifySelection A config object with an advance attribute to indicate if the selection should be moved after the link creation.
          */
-        update: function(attrs, link) {
+        update: function(attrs, link, modifySelection) {
             link = link || this.getFromSelection();
 
             if (typeof attrs === 'string') {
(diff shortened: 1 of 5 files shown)


ipeychev:

ipeychev:
Great, Chema!

It seems however there are some issues (an exception) when the word, which will be converted to a link is the last one in the text.

Thanks,

ipeychev:
Also, because we are talking about UX issues, there is a small one - when you create a link from scratch, in `componentDidMount` the test for exclusive state (`if (this.state.requestExclusive) {`) is false and the link input does not focus by default.

Thanks,

jbalsas:
Hey Iliyan,

- About the last word issue, do you mean it throws an exception, or that the toolbar stays visible? I checked and the last was the case, but maybe I changed something afterwards... in that case, anyway, we have nowhere else to go, so we'd have 3 options:
    - Keep the selection at the end of the link even if that means that the toolbar is not dismissed
    - Create an extra container to place the selection outside the link
    - Dismiss the selection completely

Since option 2 changed the user content and 3 the user cursor position, I thought the first one could be the lesser evil. What do you think we should do?

- About the UX issue, yeah, I also recently realized. I added that check so the focus will only happen when creating the link from scratch and not when updating it. I could swear it was working at some point, but I'll make sure to take a look at it tomorrow :)

ipeychev:
Hey Chema,

There aren't many options indeed. First - what do you mean with "Dismiss the selection completely"? While you explain, let's see the other options:
1. Adding an extra character will be far from the ideal. We shouldn't do this, it would be against the [code](https://en.wikipedia.org/wiki/Pirate_code).
2. To leave the toolbar as it is, will work, but honestly right now it sticks like a [cockroach on top of a cream cake](http://3.bp.blogspot.com/-P0TuYvRRELY/Th9mNgpDW3I/AAAAAAAAA8k/jhBpfe02XX4/s400/roach2.jpg).

Another possible solution is to mark the link element with some attribute on creation (data-cke-link-created) if there is no extra character after it. The toolbar will simply refuse to show if this attribute is present (and it will remove it immediately, of course). This is kinda workaround however. Let's hope your "Dismiss the selection completely" will make more sense.

Thanks,

jbalsas:
Hey Iliyan, sorry I wasn't clear enough... I meant something like what we suggested earlier with `removeAllRanges()`. But that will confuse the user as well, since focus and cursor will be lost.

The idea of `data-cke-link-created` sounds almost good to me, but I think we'll have a hard time tracking its value properly...

Do you want me to give it a try? Maybe we could pass this as it is and then work on this last issue as a bug?

What do you think?

ipeychev:
Yeah, let's try it, at least it is the harmless option from all of those.

Thanks,

jbalsas:
Hey @ipeychev,

I've pushed some changes to the PR.
- Fixed the focus issue. We were just checking the wrong property.
- In the end, I've modified the link selection test so it's never triggered if the selection is at the end of the link element. I think this is actually less invasive than what we had discussed and works fine.
- I've added tests for the advance selection feature. However, for some reason I can't get them to pass in Firefox :( The feature works, but the tests, for some reason don't reflect the selection changes properly. They pass in Chrome... what should we do? could you take a look?

ipeychev:

ipeychev:
Hey Chema,

About the advancing the cursor, OK. About tests failing on Firefox, that is true, but they fail on IE too, see the screenshot:
screen shot 2015-12-15 at 6 14 27 am

ipeychev:
Okay, on Firefox, I fixed it.
However, on IE `getSelection` returns an empty range array, that's why the exception appears.

ipeychev:
So, to make it working on Firefox, you have to call `range.optimize();` once you get the range and before doing the asserts. However, the selection on IE is just empty, damn

jbalsas:
aaarggh... `range` API is so confusing :disappointed: 

For IE, have you tried giving focus to the editor before getting the selection? I seem to remember something related to it.

I'll give it some time later today to investigate as well

ipeychev:
I did, that was the first attempt

jbalsas:
:disappointed:

ipeychev:
Hey Chema,

I don't see how to fix this on IE, so we probably should change the test or just ignore it on IE. Any ideas for changing?
Apart from that, on Firefox, when the last word is converted to a link, there is an exception, see the screencast. Would it be possible for you to take a look, please?

example

jbalsas:
Hey Iliyan, sorry I haven't had the chance to look into the IE issue so far. I'll try today, but we have the office opening and the christmas party, so it might have to be tomorrow :sweat_smile: 

I'll take a look at the Firefox exception. It seems to be coming from the `autolink` plugin, doesn't it? Why would it be triggering the `onKeyUp` method?

ipeychev:
Hey Chema,

I clicked on the button for creating link at least. I didn't have an opportunity to investigate in details where are the origins of the exception.

Good luck on the Christmas party. For curiosity, ours is today too.

jbalsas:
Hey Iliyan, I figured out the cause of the error. It also happens on Chrome, but actually only if you accept the link pressing `ENTER`.

The thing is, we create the link on `keyDown`. And we have `ae-autolink` listening on the editable area `onKeyUp`. So, we create the link, return focus to the editor, which then will catch the key up with the ENTER and trigger the logic.

The easiest fix would be to change in `button-link-edit.jsx` the `onKeyDown` atrribute for an `onKeyUp`. Do you want me to send a change with that?

I'll look into the IE issue now.

jbalsas:
Hey Iliyan, I've been trying to see how to make it work on IE, but I'm having problems to connect to the IE machines since upgrading OSX to "El Capitan". Did you update at any point? Have you experienced issues like this?

Could we ignore the tests in IE for now? I'll try to solve the issue with my vms and give it another try in the following weeks, but I'd really like to be able to push this and release `0.7.1`.

What do you think?

ipeychev:
Hey Chema,

About the Firefox fix, we may go ahead with `onKeyDown`, despite the `onKeyUp` is a better alternative. It is not fatal however and I really think resolving the issue is more important now. So, please go ahead and send this.

About the IE, did you upgrade to VirtualBox 5? If so, downgrade to 4.3.32 for example and it will start to work again. I don't have El Capitan, but I really believe the problem is in VirtualBox and `iever` we use. For me with 4.3.32 it works fine, with 5 fails.

In any case, we can ignore these two tests on IE and release 7.1 now, that makes sense. Will try to do it this weekend.

ipeychev:
Thank you, pull request merged! [See changes here.](https://github.com/liferay/alloy-editor/compare/160dc57~3...160dc57) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

Issue #420, opened 2016-01-05
Camera button throws an exception in Chrome after capturing the image


jbalsas:
Currently, the camera button image throws an exception after capturing the image. This is caused because the implementation seems to have switched to https://developer.mozilla.org/en-US/docs/Web/API/MediaStream which does not have a `stop` method anymore.

We should investigate the different ways to do it now, and the proper way to ensure it works across all the supported browsers.

antoniopol06:
I can´t reproduce this error.

ipeychev:
Hey @antoniopol06,

The issue is on Chrome, at least on my Mac. Chrome reports the following issue:
`alloy-editor-all-min.js:780 getUserMedia() no longer works on insecure origins. To use this feature, you should consider switching your application to a secure origin, such as HTTPS. See https://goo.gl/rStTGz for more details.`
In the issue description it is written: `throws an exception **after** capturing the image.`. It is actually not after, the browser refuses to create media stream at all on insecure domains.

jbalsas:
I think I saw this on Firefox after seeing that Chrome no longer worked... I tried and I can't reproduce it either, so I'll reopen if I remember what was going on :(

jbalsas:
Oh, I actually realized that the title says Chrome... wonder how did I manage to do that :joy:

ipeychev:
:)
Yeah, on Chrome. On Firefox it works like charm - no warnings and photo is there.
Here is a screencast on the live page on my machine, opened with Chrome 47.0.2526.106 (64-bit):
![bug](https://cloud.githubusercontent.com/assets/78014/12307779/c4a11c0e-ba41-11e5-93cb-d7269dc5506c.gif)

jbalsas:
Ok, after a good night's sleep, I remember how to reproduce this :relaxed: 

We just need to follow one of the two ways of [Testing Powerful Features](https://sites.google.com/a/chromium.org/dev/Home/chromium-security/deprecating-powerful-features-on-insecure-origins) in Chrome explained in the article.

Basically, run the demo on your local machine on localhost and you'll see the error.

After fixing #421, I'm guessing we won't be able to test it on chrome at all, which makes me think the fix we applied might not be the ideal one...

@antoniopol06, @ipeychev, what do you think?

ipeychev:
Hey Chema,
[...]

PR #438, opened 2016-01-20
when stream.stop doesnt exist, using stream.getVideoTracks and it stop all of them.

antoniopol06:


ipeychev:

ipeychev:
Hey @antoniopol06,

Thanks for your PR!
We have a rule how to send and accept PR in this project - there should be an issue and for this issue one or more PR can be associated. In this issue we describe the problem, and provide solution. These issues then go to the release log when we release a new version. In this case there is a PR, but no issue and because I am out of context, I don't have an idea what is the problem, on which browser does that happen and why do we need this change.
In this case I created the [issue](Chrome deprecated MediaStream.stop() method #443), please keep that in mind for the next time.

ipeychev:
Thank you, pull request merged! [See changes here.](https://github.com/liferay/alloy-editor/compare/fd3aacc~2...fd3aacc) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

antoniopol06:
Hi @ipeychev!

I wrote in the description that fixes #420 but you are right and the next time I will describe better the PR.

ipeychev:
Great! And thanks for the improvement!

PR #451, opened 2016-01-27
Fixed toolbar add loses focus and it does not appear when it is clicked.

antoniopol06:


ipeychev:

PR #458, opened 2016-02-03
Add embed support

jbalsas:
Hey @ipeychev,

This is what I'm thinking for the embed feature:
- All necessary embed plugins are provided as optional, so they're not part of the bundled files, but can be requested if the plugin is added to the `extraPlugins` list.
- I've simplified them a little bit so they don't need `NotificationAggregator`

I have a more custom/simplified version, but in the end it is the same code CKEditor plugins have and harder to get it right.

Could this be a valid approach?


ipeychev:

ipeychev:
Hey Chema,

Thanks for the contribution, this is very cool! Here we will need a bit of discussion:
1. Isn't it possible to use the required CKEditor plugins OOTB, instead to modify them?
2. Maybe we can store them to `lib/ckeditor/plugins` and they to be loaded only when added to `extraPlugins`? 
3. When pasting, the embedded stuff is occupying the whole width of the editor, but probably the creators of the plugin have thought about that? Is it possible to specify the width and height of the embedded content for example? As a user, I would expect the video to have some better aspect ratios. Or, maybe some CSS is missing? Because, on embedding a tweet I see there is a class "cke_widget_element", but I don't see any styling for it. Is it missing, or it is up to the user to apply come styling?

Thanks,

jbalsas:
Hey Iliyan,

[1-2].- That was my original idea, however, [embedbase](https://github.com/ckeditor/ckeditor-dev/blob/master/plugins/embedbase/plugin.js#L11) requires also `notificationAggregator`, and the rest of the gorilla and the jungle comes with it :joy: 

3.- That is the behaviour you can see in the [embed demo](http://sdk.ckeditor.com/samples/mediaembed.html), so I don't think we're missing any styling. I could bring this up with UX though, and apply a better default for us, what do you think?

ipeychev:
Hey Chema,

I see. There are two things which worry me:
1. We have right now `tableresize` and `tabletools`, which we get from CKEditor. Now with these we will have a few more... and I'm not sure how easy would be to update everything when we have to. Maybe we can make this migration somehow automatic or at least to document what the changes are Otherwise we will forget and we will suffer.
2. Mentioning UX, I looked on their [demo](http://sdk.ckeditor.com/samples/mediaembed.html). By default the embedded media really appears stretched (which does not make any sense IMO), but see what is written there: "_Additional widget styles are defined to control widget size and alignment._". And it is true - there is a "Narrow.." button and once you select the media, if you click on "Narrow media" and then "Center media", you will get something more useful. If we can achieve this OOTB, I think we would be good to go! We can also add these to our Styles button or we can have two buttons especially for embedded media.

Here is some other idea for discussing - do we really need all this code to embed some media? What is cool is their service which retrieves the media from an URL, and I'm not sure why they wrote all these plugins. We basically could reuse their service (ckeditor.iframe.ly/api/oembed), but I'm not sure what would be the conditions for that. The other option of course is to [host](https://github.com/itteco/iframely) the service by ourselves (Liferay to host it). Once we have that, the rest is to call a service and to add the markup. This would require more time, I guess.

Thanks,

ipeychev:
Hey Chema,

I decided to see how hard would be if we do it by ourselves, and it turned if we have the embedding service, it [won't be that hard](https://github.com/jbalsas/alloy-editor/pull/9).
What if you think if we request a machine for the embedding media? Liferay's clients will be happy to have one.
Apart from that, what is missing in this PR is:
1. To write better markup for embedding, so we can show toolbar with specific buttons.
2. To handle errors and progress.
3. Small tweaks, like better recognizing the links on paste, and of course to review the JSONP plugin - I got the first one I found.

Thanks,

PR #486, opened 2016-03-18
Add support for vertically scrollable ButtonDropdown.

harel:
Components that use this drop down can add a height property set to any value above 0.
When set, the drop down will be limited to that height in pixels and will scroll vertically if
needed.
Note that this implementation assumes the project does not use ES6 syntax.
With ES6 syntax this would look cleaner using a ...spread operator on the
properties of style and focus instead of setting them to falsly/empty values.
The onFocus was removed when scrolling because in a scrolling situation this
caused any selection beyond the visible area (i.e., when having to scroll down)
to break down and select the first option again and again

### DIFF
--- a/src/ui/react/src/components/buttons/button-dropdown.jsx
+++ b/src/ui/react/src/components/buttons/button-dropdown.jsx
@@ -17,6 +17,7 @@
          */
         getDefaultProps: function() {
             return {
+                height: 0,
                 circular: false,
                 descendants: '.ae-toolbar-element',
                 keys: {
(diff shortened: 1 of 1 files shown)


ipeychev:

ipeychev:
Hey @harel,

Thanks for your PR. Styling we usually handle using CSS and in this particular case, it is totally doable to do the following:
```css
.ae-dropdown {
    overflow-y: scroll;
    max-height: 100px;
}
```
and it will make the rest. The question is: do you see some benefits in introducing another property instead using CSS?

About the focus issue, such indeed exist (reproducible when the above CSS is applied too) and for that reason let's /cc @jbalsas too.

Thanks,

harel:
The class name solution is fine,  better than styling directly as it goes against the code style of the project - I get it. There reason I did it with a prop was to allow drop downs that can scroll along size drop downs that don't need scrolling in the same toolbar. If the class name is applied to all drop downs will it not prevent such a scenario?

ipeychev:
Yeah, it will, but normally you would like a consistency and all of the dropdowns to have max X pixels height. If not, targeting a specific dropdown might be achieved with a bit more complex CSS.
Also, we may think to introduce a common property for all components, which will allow the developer to apply custom class.

The main issue here is the interface - the developer passes `height`, and we [translate](https://github.com/liferay/alloy-editor/pull/486/commits/0871ba6b43fcc369d6343a0f0de360741ce4fdbd#diff-cfa7bf250d770857b2e7f531b12a505cR55) it to some styles. This might be fragile - what if these styles are not enough for some browsers in the future? The developer should open an issue and wait until we fix it. Using CSS he can do it immediately. Second, tomorrow we might want to apply some other stylings in the dropdowns (or other components) - `width` for example. Then, we have to add another property, it should be translated to other styles and so on. Using CSS we allow people to style whatever they want without touching the code.

Does that make sense?

Thanks,

harel:
Yes it makes sense. A custom class property will solve it when its available as we'll be able to do it using the class. 
In which case - this PR can be closed. At least we found the focus issue :)

ipeychev:
Yeah, and thanks for the contribution!

Issue #120, opened 2015-03-11
Add autocomplete to links


ipeychev:

kushal:
Do you meant from a server provided list of links? FWIW, I have a local fork that does this which I'd be happy to try to merge, but I'm not sure how exactly you'd want me to inject the completions. If you had an outline of how you wanted this to work, I could try to move in that direction.

ipeychev:
Well, currently there is an input field where the user can type the link URL.
One option is to make this a standard autocomplete component which supports server and client datasource. The buttons accept configurations, so the user will be able to configure the datasource via ButtonLinkEdit's configuration. In most cases, this will be the URL, which will return the results, or just an array with the data to be filtered in the client.

PR #488, opened 2016-03-21
Link autocomplete

mattleff:
This pull aims to add support for link autocompletion (#120) using a remote API. It looks something like this:

image

### Known issues

- [x] XHR should not be in presentational component
- [x] Close target dropdown when autocomplete dropdown is open and vice versa
- [ ] `autocompleteUrl` prop is not set on link edit, which causes autocomplete to not function
- [ ] Tests

I'd like some feedback on whether the `autocompleteUrl` string prop is a better loading strategy than doing a function prop which would put the XHR responsibilities on the consumer rather than in core. This:

```js
{
	name: 'link'
	cfg: {
		autocompleteUrl: 'action/article/title/search?term={term}'
	}
}
```

versus something like this:

```js
{
	name: 'link'
	cfg: {
		autocompleteUrl: function(term) {
			var that = this;
			$.get( 'action/article/title/search?term=' + term, function(data) {
				that.setAutocompleteOptions( data.links );
			}
		}
	}
}
```


ipeychev:

ipeychev:
Hey @mattleff,

Thanks for your contribution! About `autocompleteURL` being string or function:
In an autocomplete the most powerful approach is to have one property, called for example "data", which could be one of these:
- array - the data itself
- function - this function will be in charge of retrieving the data. If a function, the query(term) will be passed to it and it is supposed to return a Promise, which then should be resolved with the data.

We can also support a string, and to replace "term" or "query" inside, which will require the editor to make the XHR request itself. In a client application there is always some way to deal with XHR, so if we implement our own XHR, it might be just overhead.

How does that sound for you?

Thank,

mattleff:
@ipeychev Thanks for the review! I think the `data` idea is good, and I think just supporting a data array or a data function is probably the best solution. Since some (most?) won't use the autocomplete functionality not including the XHR code will help reduce the code size of this feature. I'll see if I can implement the array/Promise solution.

Is there a "system" way to deal with closing the target dropdown when autocomplete dropdown is open and vice versa, sort of like exclusive dropdown?

Also, did something I commit break the tests or are they not working right now?

ipeychev:
Hey @mattleff,

About the tests - they are working, and there is nothing broken. If you mean the error [here](https://travis-ci.org/liferay/alloy-editor/builds/117599307), for some reason Travis couldn't run the browsers on SauceLabs. However, on my machine they all pass, which is the important part. I trust my machine much more than I trust Travis and SauceLabs ;)

Thanks,

ipeychev:
Umm, no there is no currently some kind of "Exclusive dropdown". Let me ping @jbalsas, maybe he will have some idea?

Thanks,

jbalsas:
Hey guys, sorry I'm a bit off the grid.

@mattleff Thanks a lot for the contribution. It looks awesome!! It is something that's been on our roadmap for a while but never had time for it.

@ipeychev, @mattleff, the dropdowns are exclusive by design. Take a look at the table selection in http://alloyeditor.com/demo/ and you'll see how only one is expanded at a time. Maybe I did something wrong with the link target one so that this is not the case. Is it happening in this PR? I can take a look later today to try to figure out what's going on.

ipeychev:
Hey Chema, sure, they do. However, [here](https://github.com/liferay/alloy-editor/pull/488/commits/401073a1029c5123c2ef2f3c1d49ad65044847a8#diff-c4ff6c3d86109cc3fc9e5b4345dc44d1R169), when the `state` is set, both dropdowns become visible. From that I assumed @mattleff is asking about a way to _explicitly_ say to the dropdowns to show/hide.

Btw, @mattleff , I modified the code a bit in order to see it in action, because I didn't have a server, which would return a set of items. Having `data` property which supports both simple array and a function would be wonderful.

Thanks,

jbalsas:
Yeah, I see what you mean. It should be automatic if the component uses the injected `this.props.toggleDropdown` method.

The way to render the dropdown should be using the `this.props.expanded` prop, and then use `this.props.toggleDropdown` to make the dropdown appear.

mattleff:
@ipeychev Thanks for the info on the tests. I tried running them locally and I got the same error as the CI so I wasn't sure.

On the `data` property: Are you thinking to have the data array be filtered to the `term` by the component or simply show all the options in the array?

@jbalsas Thanks for the info on `this.props.expanded` and `this.props.toggleDropdown`. I'll see if I can get that working.

ipeychev:
@mattleff About `data` property: term will be always used to filter the data. The only difference is where this data comes from - it might be directly passed as an array, or it could be retrieved from the server (the function will take care of that). In both cases, a Promise could be used and the setter of `data` property will normalize this.

mattleff:
@ipeychev 

>  ...term will be always used to filter the data.

True, but I was picturing that in the Promise case we'd assume the data was pre-filtered, meaning we'd just list the items returned. That way if I type `Bananas` my API could return an array element that wouldn't have `Bananas` in either the title or the url. Is that what you're thinking?

In the data array case our only option is to filter the title (or url?) by the term.

ipeychev:
Here is what I mean:
1. We create a property, like this:
```
data: '_setData'
```
In the `_setData` setter, if the passed value is not a function, then we wrap it in a function, which will return the data.
2. Once we have the `term` and we are ready to retrieve the data, we pass it to this function as a parameter. If the returned result is not a Promise, we make it a Promise using [Promise.resolve](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve) method.
3. Once the Promise resolves, we process the returned value and if it is an Array, we populate it.

Does it make sense?

Thanks,

mattleff:
@ipeychev Thanks, that helped! See what you think of a5a8105.

ipeychev:

ipeychev:
Yeah, great! Will take deeper look this weekend and merge it :)

Thanks again!

ipeychev:
After some additional checking, other issues appeared:
1. In edit mode, the autocomplete does not appear at all. This is because we pass the `data` to button `link` which in turn propagates it down to `linkEdit`. However, when there is an already existing link, we directly render `linkEdit` and in this case `this.props.data` is undefined. We have to think how to fix this.
2. The keyboard support is missing - from linkEdit you have to press Tab in order to move to the dropdown items. An autocomplete cannot work without this. Chema?

Thanks,

jbalsas:
Hey guys, I'll take a look at this... `focusManager` should be able to handle most of the keyboard support we need.

ipeychev:
Very good, Chema. There is another [issue](https://github.com/liferay/alloy-editor/issues/491) with the focusManager.

Thanks,

ipeychev:
Hey @mattlef/@jbalsas,

Instead to pass the data to ButtonLink and propagate it down to ButtonLinkEdit, which as it looks doesn't work in all situations, another option for us might be to create a plugin, something like "LinkAutoComplete" and configure it from the editor's config. We will pass the data source and it will add the autocomplete functionality to  ButtonLinkEdit.

How does that sound for you?

Thanks,

ipeychev:
Hey guys, something else came to my mind:
The other option is to leave the things as they are in this PR and pass the autocomplete data from another property in the config. In this case, the only change would be that ButtonLinkEdit will get the data from the `data` property, if passed, or from the editor's config via `getDefaultProps` method.
When we are creating tests, we will pass the `data` directly, when the user is configuring the editor, he will pass the data in the config from another property, something like `linkDataSource`, or `linkData`.

If that sounds OK for you, I can add this functionality really quick and merge this PR. In any case however we shall resolve the keyboard issue.

@mattleff @jbalsas

mattleff:
@ipeychev I'm good with the `linkData` idea. The only downside of an editor config property is that you'd need to provide the same data for all link buttons if you had a link button in more than one toolbar. I doubt that's an issue for anybody so I can't think of any downside. :+1:

jbalsas:
Hey guys, I think we do something like that for some of the
button-link-target props, since it is nested the same way...

If you guys are happy with the pr, go ahead and merge. I'll try to work on
the keyboard and focus issues asap (although it might take some time until
I get there, sorry :( )

Thanks!

ipeychev:
Hey Chema,

About [button-link-edit](https://github.com/liferay/alloy-editor/blob/492fbecc45de35cf53be72cf7aefb3abacb00593/src/ui/react/src/components/buttons/button-link-edit.jsx) you mean the code [here](https://github.com/liferay/alloy-editor/blob/492fbecc45de35cf53be72cf7aefb3abacb00593/src/ui/react/src/components/buttons/button-link-edit.jsx#L211), don't you? That's an interesting case, because I think it would have the same issue - it receives `allowedLinkTargets` or provides default ones, but how exactly they could be propagated from the configuration? They can, but when the ButtonLinkEdit is instantiated [explicitly](https://github.com/liferay/alloy-editor/blob/492fbecc45de35cf53be72cf7aefb3abacb00593/src/ui/react/src/selections/selections.js#L10), not via ButtonLink, the default ones will be provided and not those, specified by the user.
@jbalsas  

Matthew, you mean that if we provide the data from the configuration, _all_ button links would retrieve the same data from this property, don't you? That is true and a potential issue here would be how the developer would pass _different_ data to the different link edit buttons, if needed.

To visualize and paste some code this would look like this:
```
AlloyEditor.editable('editable', {
    linkData: [{
        title: 'Matthew',
        url: 'http://matthew.com'
    }, {
        title: 'Chema',
        url: 'http://chema.com'
    }]
});
```
and then _all_ buttons for editing links would get same data. I personally don't think this would be an issue for this button specially, but it might be for some buttons in the future. I really would love to propagate the data [here](https://github.com/liferay/alloy-editor/blob/492fbecc45de35cf53be72cf7aefb3abacb00593/src/ui/react/src/selections/selections.js#L10) too, but I don't see how.

There is something else - if we expose the property `linkData`, tomorrow we may have to provide another configuration for ButtonLinkEdit, or for some other button. I'm afraid that we may end up with multiple floating properties, which apply to different buttons, like:
```
AlloyEditor.editable('editable', {
    linkData: [{
        title: 'Matthew',
        url: 'http://matthew.com'
    }, {
        title: 'Chema',
        url: 'http://chema.com'
    }],
    linkTargets: [{
        label: AlloyEditor.Strings.linkTargetSelf,
        value: '_self'
    }, {
        label: AlloyEditor.Strings.linkTargetBlank,
        value: '_blank'
    }, {
        label: AlloyEditor.Strings.linkTargetParent,
        value: '_parent'
    }, {
        label: AlloyEditor.Strings.linkTargetTop,
        value: '_top'
    }]
});
```

And this is close to the chaos.

Another option might be to expose one property per button, where all its configuration would be. Of course, this applies to buttons, which are _hidden_ in the configuration (the developer does not configure them explicitly, like buttonLinkEdit). This may look like this:
```
var alloyEditor = AlloyEditor.editable('editable', {
    toolbars: [...],
    buttonCfg: [
        'linkEdit': {
            data: [{
                title: 'Matthew',
                url: 'http://matthew.com'
            }, {
                title: 'Chema',
                url: 'http://chema.com'
            }]
        },
        'linkTargetEdit': {
             allowedTargets: [{
                label: AlloyEditor.Strings.linkTargetSelf,
                value: '_self'
            }, {
                label: AlloyEditor.Strings.linkTargetBlank,
                value: '_blank'
            }, {
                label: AlloyEditor.Strings.linkTargetParent,
                value: '_parent'
            }, {
                label: AlloyEditor.Strings.linkTargetTop,
                value: '_top'
            }]
         }
    ]
})
```

The last option is the most powerful one - the developer will be able to configure the buttons, which don't appear explicitly in the toolbars, but are instantiated from other buttons.

Sorry from the long post, but I hope you will read it carefully and give some feedback. This is really important, because we have an interface issue and we have to resolve it in a smart way.

Thanks,

@mattleff, @jbalsas

mattleff:
@ipeychev Yeah, that's the issue I was wondering about. As you say, it's probably not an issue for this use case, but it's good to think through the interface/precedent.

Of the options you listed I lean toward something like the last option. It didn't really make sense to me that I needed to override the selections in order to pass data to a button. However if we need to support the case where a given button will have multiple different sets of configuration data (from being in multiple toolbars) I think passing the data in the selections is the only way it can work. Perhaps we can make the common use case simple.

Wdyt of something like this:

```
var alloyEditor = AlloyEditor.editable('editable', {
    toolbars: [...],
    buttonCfg: {
        'linkData': [{
            title: 'Matthew',
            url: 'http://matthew.com'
        }, {
            title: 'Chema',
            url: 'http://chema.com'
        }],
        'linkAllowedTargets': [{
            label: AlloyEditor.Strings.linkTargetSelf,
            value: '_self'
        }, {
            label: AlloyEditor.Strings.linkTargetBlank,
            value: '_blank'
        }, {
            label: AlloyEditor.Strings.linkTargetParent,
            value: '_parent'
        }, {
            label: AlloyEditor.Strings.linkTargetTop,
            value: '_top'
        }]
    }
})
```

I think this kind of global object has some precedent with the way CKEditor handles config options. CKEditor uses [`CKEDITOR.config`](http://docs.ckeditor.com/#!/api/CKEDITOR.config) and `editor.config` to store global settings (aside: it would be nice to be able to change AlloyEditor configs for all instances with a single call, like: `AlloyEditor.config.buttonCfg.linkData = [...]`.)

The first downside I can think of is that the config options aren't *automagically* passed to the appropriate button. The button would need to be able to read them out of the global config object. If we did your last option we could pass only the config options that were relevant to a given button. However, I think this would break down quickly if there ever needed to be a setting that would affect multiple buttons (for example, something like `forceProtocol` which would make all buttons generate their resource tags with `https`. This setting would need to be available for all buttons...).

Thoughts? @ipeychev @jbalsas

ipeychev:
Hey @mattleff,

I would go with Option 3) from the suggestions above, because in this case we would pass the configuration to each button explicitly and we have the advantage that all the configurations are grouped per button.
If something is applied to multiple buttons, like `forceProtocol`, it should be a general configuration property. To remember, here it how it will look like:
```javascript
var alloyEditor = AlloyEditor.editable('editable', {
    toolbars: [...],
    buttonCfg: [
        'linkEdit': {
            data: [{
                title: 'Matthew',
                url: 'http://matthew.com'
            }, {
                title: 'Chema',
                url: 'http://chema.com'
            }]
        },
        'linkTargetEdit': {
             allowedTargets: [{
                label: AlloyEditor.Strings.linkTargetSelf,
                value: '_self'
            }, {
                label: AlloyEditor.Strings.linkTargetBlank,
                value: '_blank'
            }, {
                label: AlloyEditor.Strings.linkTargetParent,
                value: '_parent'
            }, {
                label: AlloyEditor.Strings.linkTargetTop,
                value: '_top'
            }]
         }
    ]
});
```
In the above code, 'linkTargetEdit' and 'linkEdit' are the buttons.

Let's see what @jbalsas will say and then we can go and implement this.

ipeychev:
Okay, guys [here](https://github.com/liferay/alloy-editor/commit/fe4ea3c1b305cdb8d91df801bf129d65f0b68a62) is the committed code, following what we've discussed so far. Could you please check if that is OK for you? If not - we go with another iteration.

Also, @mattleff's would you be so kind to update your commit and pass the data not via `ButtonLink`, but directly to `ButtonLinkEdit`? Let us know how convenient is that for the developer and if you have some other suggestions. After that we merge your PR.

Thanks,

jbalsas:
Hey @ipeychev, @mattleff, figuring out a way to easily configure stuff has always been a recurrent topic for us. I think this approach helps for now, so I'm fine with it.

In my mind, something more long-term I'd like to have would be some kind of self-generated configuration DSL. Something where you could do:

```javascript
let linkEditButtonCfg = configDSL.getButton('linkEdit');
    linkEditButtonCfg.setData({...});
    linkEditButtonCfg.setAllowedTargets([...]);

let linkSelectionCfg = configDSL.getSelection('link');
    linkSelectionCfg.setTestFn(() => {...});
    linkSelectionCfg.addButton(linkEditButtonCfg);

...
```

If we could somehow extract from the props annotations what's available for each button configuration and expose it as an API that easily allowed people to discover and build on top of that, that could ease the pain on generating complex and custom configurations.

I'm really sorry I don't have a lot of time at the moment to dig deeper either way (been in Brazil for a couple of weeks and I'm now in L.A. for another couple :cry: ). As I said, I'm fine with this going forward, I'll get back to it as soon as I get some time again so we can iterate if needed.

Thanks for all the work!

ipeychev:
Thank you, pull request merged! [See changes here.](https://github.com/liferay/alloy-editor/compare/fbafd4b~3...fbafd4b) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

ipeychev:
Hey @mattleff,

You PR has been merged. Thanks for your contribution!
In addition, a new mixin has been added in order to merge the button props with user's configuration from `buttonCfg` property. Starting from now, we should be able to pass a `data` function to button `linkEdit` which will retrieve the data from the server according to the term. The code should look like this:
```javascript
    buttonCfg: {
        'linkEdit': {
            data: function(term) {
                return new Promise(function(resolve, reject) {
                    // retrieve data from server according to the term
                    // then resolve the Promise
                });
            }
        }
    }
});
```

@jbalsas, take your time and when you have a chance, please take a look on the issue with keyboard support for autocomplete.

Thanks,

mattleff:
@ipeychev Looks great! Thanks for all your help!

PR #501, opened 2016-04-18
Lang autogenerated

antoniopol06:


src/ui/react/src/selections/selections.js
```diff
@@ -16,7 +16,7 @@
         test: AlloyEditor.SelectionTest.image
     }, {
         name: 'text',
-        buttons: ['indentBlock'],
+        buttons: ['styles', 'bold', 'italic', 'underline', 'indentBlock', 'outdentBlock', 'link', 'twitter'],
```
jbalsas: I wouldn't add these by default. We need to create the feature videos and add them to the site so people can find them, but let's keep the toolbar simple and not force people to always strip stuff they don't need.

Issue #517, opened 2016-05-12
Cannot Remove Embed Link from Editor


btknorr:
We are using the most basic form of AlloyEditor, no plugins or anything.  Just using AlloyEditor.editable('myContentEditable'); like stated in the guide.

Everything works great until you add an embed link.  Once it's there, it cannot be removed.  How do you remove an embed link?  If you select it, the embed link field pops up, and if you click the X, and then the checkmark, it still remains.  If you select it and then hit the backspace key, the browser goes back in it's history to the previous web page.  Very frustrating...please advise.

Thanks,

Brian

btknorr:
Try it yourself using the AlloyEditor demo, just add an embed link and then try to remove it :)

http://alloyeditor.com/demo

jbalsas:
Hey @btknorr, I just verified the issue. I think it only happens if the embed content has an iframe. We have some code that forces the widget selection so it triggers a `selectionChange` event which seems to be causing the inner iframe to be focused making it impossible to remove.

If you're building AlloyEditor yourself, you may try commenting [this line](https://github.com/liferay/alloy-editor/blob/master/src/plugins/embed.js#L104) and rebuilding the editor.

We'll work on this next week to come up with a complete solution to this.

Thanks for letting us know about it, and sorry for the inconvenience!

jbalsas:
On a related note, this fix in upstream CKEditor might be relevant: http://dev.ckeditor.com/ticket/14538

ipeychev:
Hey Chema,

I was going to release this weekend ver. 1.5.1 which includes the fix for https://github.com/liferay/alloy-editor/issues/515 (great job from @antoniopol06 and you), but in this case let's postpone it for one week and include the fix for this issue too, okay?

Thanks,

jbalsas:
Hey Iliyan, yeah, I agree. We'll work on this next week. 

BTW, I think we should be planning `1.2.1`, we're currently at `1.2.0` 😅

ipeychev:
Yeah, correct, `1.2.1` of course - too many projects and I started to get confused which one is on which version :D 
Btw, talking about versioning, do you know [Semantic release](https://github.com/semantic-release/semantic-release)/[Commitizen](https://www.npmjs.com/package/commitizen)? They help people to have better change logs and better versioning. Maybe it will be good idea to think for adopting them.

Thanks,

jbalsas:
I just sent a fix proposal for this (#517), including:
[...]

PR #524, opened 2016-05-21
Improve ae-embed plugin widget selection

jbalsas:
Hey @ipeychev, this is a possible fix for #517.

We've changed the way the widgets are selected and handled to let CKEditor handle it in its own way, and figured out a different way to show the toolbar on them.

On top of that, we've also added a remove button to the embed edit button following the pattern on the link one so it is possible to remove a widget from the toolbar.

Thanks!


ipeychev:

ipeychev:
Hey Chema,

This is really cool, but right now clicking on the button for adding an embedded media throws an exception? Please see the attached screencast.

Thanks,
bug

jbalsas:
Oh! Forgot to check that one! I was always pasting a url

I'm not at home right now, but I'll take a look as soon as a I get back!

ipeychev:
Great, the rest is cool!

One thing to consider - maybe it will make sense to allow people to set the size of the embedded content. Right now it takes the whole content width and there is no way to change it. In the same time, I'm not sure if we can make it perfect for every provider.
As I see, the response contains the name of the provider - "provider_name": "YouTube" or "provider_name": "Twitter" for example. Maybe we can add such class to the wrapping div, something like `data-provider=YouTube`, so people will be able to style it accordingly? 

Thanks,

jbalsas:
Hey Iliyan!

I've pushed one extra commit with some fixes for the remove button. I've refactored it so it better follows the same pattern as `button-link-edit`.

This has also made me realize we never got around to create tests for this feature, so I'll ask @antoniopol06 to work on that whenever we can.

jbalsas:
Regarding the width and height of the embeds, I remembered you already brought this up on our initial discussion and I never followed on this. I'm really sorry about that 😢 

How about we create an independent feature request for it so we can properly analyze it and tackle it as soon as we can?

I get the feeling that just a class might not be enough. Depending on the media content, it might be necessary to, for example, set the `width` and `height` attributes of the iframe, or maybe even request the embed again if the query supports dimensions as the output might be different.

ipeychev:

ipeychev:
Thank you, pull request merged! [See changes here.](https://github.com/liferay/alloy-editor/compare/1f354f4~2...1f354f4) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

ipeychev:
Hey Chema and Antonio,

Great, this was cool fix which means it is time to release another minor version!

About setting the size of the embedded content, if you think the class won't be enough, then, sure - let's open a ticket to consider what can be done.

@jbalsas @antoniopol06

PR #530, opened 2016-05-26
Fixes #529

### LINKED ISSUE #529: Selection menu doesn't display when selecting ends outside editable
### Steps

1. Visit http://alloyeditor.com/demo/
2. Select `Paste Content`, highlighting from right to left and dragging the mouse outside `#contentEditable`.

### Expected result

The selection menu should be shown.

### Actual result

The selection menu is not shown. If you select from left to right the menu is shown.

Pull #530.

mattleff:


mattleff:
[`setCapture`](https://developer.mozilla.org/en/docs/Web/API/Element/setCapture) may also be a possible fix, but that might require polyfilling for Alloy's supported browsers and it seems to have some issues of its own.

ipeychev:

ipeychev:
Hey @mattleff,

Thanks for the contribution! The event "mouseout" is very expensive - it is not fired only when the user is leaving the content editable area. It will be invoked when the mouse is over any child element.
To check that, you may replace the code in the patch with this one:
```javascript
editable.attachListener(editable, 'mouseout', function(event) {
    console.log('MOUSEOUT');

    handleUI(event);
});
```
and move the mouse over links, the toolbars, etc. You will see the how many times the listener is invoked.

What we probably need is `mouseleave` event. Look [here](https://developer.mozilla.org/en-US/docs/Web/Events/mouseout) for more information. However, mouseleave will be invoked even when we move the mouse over a Toolbar (because they are rendered outside the editable area), which means if we just replace `mouseout` with `mouseleave`, there will be some movement of styles toolbar. 
Maybe the right solution should be like this:
1. We replace `mouseout` with `mouseleave`. We might have to debounce the listener, but initially we may try even without it.
2. Inside listener, we check if the mouse is on top of some of our Toolbars. If it is, we don't do anything, otherwise, we call `handleUI`.

Do you think you can do it?

Thanks,

mattleff:
Hey @ipeychev, all good points. Let me see what I can come up with.

mattleff:
@ipeychev Let me know what you think of what I just pushed. I didn't debounce it, but I could add that if you think we need it.

ipeychev:

ipeychev:

ipeychev:
Thank you, pull request merged! [See changes here.](https://github.com/liferay/alloy-editor/compare/7e3c66d~2...7e3c66d) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

ipeychev:
Hey @mattleff,

I accepted the PR with some changes, you may take a look on [them](https://github.com/liferay/alloy-editor/commit/f57ef315b0154543238e20adf7453a28d75a0a61). I added a short description of them in the commit message too. Please let me know if everything is clear.

Thanks,

mattleff:
@ipeychev Looks great, thanks for the commit!

PR #537, opened 2016-06-07
Embed tests (Fixes #518)

### LINKED ISSUE #518: Multi-image drag-n-drop trigger unique event
Currently dragging-n-dropping and pasting an image trigger the same event ([here](https://github.com/liferay/alloy-editor/blob/master/src/plugins/addimages.js#L180) and [here](https://github.com/liferay/alloy-editor/blob/master/src/plugins/pasteimages.js#L71)). While this is good for most cases I'm wanting to have a separate event for a multi-image drag-n-drop. This allows me to turn several images into a single element in the editor (i.e., for a gallery).

I've added this to our repo with [this](https://github.com/SimpleUpdates/alloy-editor/blob/multi-image-event/src/plugins/addimages.js#L65-L88) but I was wondering if this is something that you would want or be willing to have. Or is there a better way to accomplish what I'm doing with the current code?

jbalsas:
Hey @ipeychev, this adds tests for the embed functionality (#518).

Just as a curiosity, this improves test coverage the following way:

Metric | Before | After
------------ | ------------ | -------------
_Statements_ | 71.0% (2204/3101) | 73.11% (2267/3101) 
_Branches_ | 55.76% (895/1605) | 57.76% (927/1605)
_Functions_ | 74.85% (488/652) | 77.76% (507/652)
_Lines_ | 71.49% (2177/3045) | 73.56% (2240/3045)

We'll try to keep adding tests for the missing buttons and features over the next few weeks.

PS: I just switched to a new computer, so I'm right now installing the VMs to run the IE tests... just sending this in case you can test it sooner 😉


jbalsas:
There are actually test failures in IE, so closing and resending...

Issue #532, opened 2016-05-31
can't remove the marker


myblackboxgithub:
I would like to make AlloyEditor to switchable between enable and disable.
When a imaged is selected and destroy editor, it is fine, but when it back to enable, it remains ‘ckimgrsz’ class, margin and anchor are still there, and it makes error messages, 'lock' of null or 'unlock' of null.

how can I remove it? thanks

`var myEditor;
//to enable
if (!CKEDITOR.instances.introduction)
	myEditor = AlloyEditor.editable('introduction');

//to disable
if (CKEDITOR.instances.introduction)
	myEditor.destroy();`

![bugs](https://cloud.githubusercontent.com/assets/15113714/15672912/59abfcf0-272b-11e6-8b5f-18ad934b90e7.gif)

jbalsas:
Hey @myblackboxgithub, we'll take a look into this... I agree this probably shouldn't happen, and it might be as easy as properly clean up within the `imageresize` plugin.

In the meantime, we implemented a less expensive way of disabling the editor using [`.setReadOnly`](https://github.com/liferay/alloy-editor/issues/436), so you might give it a try to see if that somehow changes how this behaves.

myblackboxgithub:
thanks @jbalsas,
I tried it with .setReadOnly, it seems work fine, but the selected mark still remains and it can be resizable image. 

then I added some code what I found Internet to make it unselect. it seems works fine. 
thanks for help.

`function enableEditing() {
    if (!CKEDITOR.instances.introduction) {
        myEditor = AlloyEditor.editable('introduction');
    }
    else {
        myEditor.get('nativeEditor').setReadOnly(false);
    }
}

function disableEditing() {
    if (CKEDITOR.instances.introduction) {

//to remove selection
        var oEditor = CKEDITOR.instances.introduction;
        var doc = oEditor.document,
            range = new CKEDITOR.dom.range(doc),
            body = doc.getBody();
        range.setStart(body, 0);
        range.setEnd(body, 0);
        oEditor.getSelection().selectRanges([range]);
        oEditor.getSelection().removeAllRanges();        

        myEditor.get('nativeEditor').setReadOnly(true);
    }
}`

jbalsas:
Hey @myblackboxgithub, glad it's working for you now!

[...]

PR #543, opened 2016-06-08
Clear browser selection on editor destroy (#532)

jbalsas:
Hey @ipeychev, this is a possible fix for #532.

Directly invoking CKEditor's `editor.getSelection().removeAllRanges()` won't generally work since the method usually does nothing when the editor is not focused (and focusing it prior to destroying it causes some issues), so we tried to do it manually.

What do you think?


ipeychev:

ipeychev:
Great!

PR #630, opened 2016-11-18
Add style to dropwdown when element was selected

antoniopol06:


src/ui/react/src/uibridge/richcombo.jsx
```diff
@@ -137,12 +137,15 @@
                         var richCombo = this;
 
                         var items = this._items.map(function(item) {
+
+                            var className = 'ae-toolbar-element ' + (item.value === this.state.value ? 'ae-toolbar-element active' : '');
```
jbalsas: There's an extra `ae-toolbar-element` here

antoniopol06:
Ok! I will add to them and I will resent the PR.

Thanks!

PR #640, opened 2016-11-25
dragresize plugin condition is moved to initializer to avoid off breaking change.

antoniopol06:


src/ui/react/src/adapter/alloy-editor.js
```diff
@@ -43,6 +45,12 @@
 
             editor.config.selectionKeystrokes = this.get('selectionKeystrokes');
 
+            if (config.extraPlugins) {
```
jbalsas: Hey @antoniopol06, I think this is not 100% right. This will always add the drag_resize plugin to the `extraPlugins` property, even if the user has decided to have a different set of plugins.

We should add `ae_dragresize` to the default value and simply replace the plugin with the ie version here in case it appears in the property on initialization.

ipeychev:

ipeychev:
Thank you, pull request merged! [See changes here.](https://github.com/liferay/alloy-editor/compare/e3b182b~1...e3b182b) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

ipeychev:
Hey @antoniopol06,

Please keep in mind the commits should be reverted using `git revert` command, or if you do it manually, the commit message should clearly say which commit is being reverted.

Thanks,

PR #648, opened 2016-12-02
Test placeholder is removed when editor is changed and it has content

antoniopol06:


src/plugins/placeholder.js
```diff
@@ -32,6 +32,7 @@
              */
             init: function(editor) {
                 editor.on('blur', this._checkEmptyData, this);
+                editor.on('change', this._checkEmptyData, this);
```
jbalsas: We probably want to debounce this, don't we?

ipeychev:

src/plugins/placeholder.js
ipeychev: However, after debouncing the test will fail.

ipeychev:
Thank you, pull request merged! [See changes here.](https://github.com/liferay/alloy-editor/compare/25be6c3~3...25be6c3) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

PR #683, opened 2017-01-26
Adds a plugin discovery feature

### LINKED ISSUE #667: Allow discovery of plugin buttons via `ae_uibridge`
AlloyEditor supports [using CKEditor plugins](http://alloyeditor.com/guides/how_to_use_ckeditor_plugins/) thanks to the `ae_*bridge` set of plugins. However, in order to add the buttons to a toolbar, the user needs to figure out how is the particular plugin naming the button, so for instance, in order to use the `Paste from Word` plugin, one would need to do the following:

```js
    extraPlugins: AlloyEditor.Core.ATTRS.extraPlugins.value +',ae_uibridge,ae_buttonbridge,pastefromword,clipboard,dialog,dialogui',
    toolbars: {
        add: {
            buttons: ['table', 'hline', 'image', 'camera', 'PasteFromWord']
        },
        ...
    }
});
```

Since this is error-prone and forces the user to dig into the plugin code for the exact string, it would be nice to have some kind of API so a user could simply do:

```js
    extraPlugins: AlloyEditor.Core.ATTRS.extraPlugins.value +',ae_uibridge,ae_buttonbridge,pastefromword,clipboard,dialog,dialogui',
    toolbars: {
        add: {
            buttons: ['table', 'hline', 'image', 'camera', ...AlloyEditor.getButtonsFor('pastefromword')]
        },
        ...
    }
});
```

julien:
Fixes #667

### DIFF
--- a/src/ui/react/src/uibridge/richcombo.jsx
+++ b/src/ui/react/src/uibridge/richcombo.jsx
@@ -223,7 +223,7 @@
 
                 var e = new Error();
                 // In IE 9 - IE 11, e = {number: 0, description: ''};
-                // so check the "stak" property
+                // so check the "stack" property
                 if (!e.stack) {
                     return;
                 }
--- a/src/ui/react/src/uibridge/richcombo.jsx
+++ b/src/ui/react/src/uibridge/richcombo.jsx
@@ -219,7 +219,20 @@
          */
         init: function(editor) {
             editor.ui.addRichCombo = function(richComboName, richComboDefinition) {
-               this.add(richComboName, CKEDITOR.UI_RICHCOMBO, richComboDefinition);
+                this.add(richComboName, CKEDITOR.UI_RICHCOMBO, richComboDefinition);
+
+                var e = new Error();
+                // In IE 9 - IE 11, e = {number: 0, description: ''};
+                // so check the "stak" property
+                if (!e.stack) {
+                    return;
+                }
+
+                var pluginName = /plugins\/(.*)\/plugin.js/.exec(e.stack)[1];
+                if (!AlloyEditor.BRIDGE_BUTTONS[pluginName]) {
+                    AlloyEditor.BRIDGE_BUTTONS[pluginName] = [];
+                }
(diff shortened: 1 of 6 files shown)


jbalsas:
Hey @julien, thanks for this. I'll take a look as soon as possible.

src/ui/react/src/uibridge/richcombo.jsx
```diff
@@ -219,7 +219,20 @@
...
+                this.add(richComboName, CKEDITOR.UI_RICHCOMBO, richComboDefinition);
+
+                var e = new Error();
+                // In IE 9 - IE 11, e = {number: 0, description: ''};
+                // so check the "stak" property
+                if (!e.stack) {
+                    return;
+                }
+
+                var pluginName = /plugins\/(.*)\/plugin.js/.exec(e.stack)[1];
```
jbalsas: Hey @ipeychev, what do you think about this? 

It is the only way we could figure to actually map the plugin names. This will happen rarely, so we can take the performance hit. The only problem I see is that in a bundle scenario, this might not work... In light of that, we may not be able to implement this in a safe and consistent way... what do you think?

jbalsas: @ipeychev, thoughts on this?

src/ui/react/src/uibridge/richcombo.jsx
```diff
@@ -219,7 +219,20 @@
          */
         init: function(editor) {
             editor.ui.addRichCombo = function(richComboName, richComboDefinition) {
-               this.add(richComboName, CKEDITOR.UI_RICHCOMBO, richComboDefinition);
+                this.add(richComboName, CKEDITOR.UI_RICHCOMBO, richComboDefinition);
+
+                var e = new Error();
+                // In IE 9 - IE 11, e = {number: 0, description: ''};
+                // so check the "stak" property
```
jbalsas: Typo: `stak` should be `stack`

julien: thanks

PR #695, opened 2017-02-09
Added padding to td and th. These elements are clickable when in the first column was introduced a huge text.

antoniopol06:


jbalsas:

PR #703, opened 2017-03-06
Hash autogenerated by build-languages task

antoniopol06:


ipeychev:
@jbalsas, was that PR reviewed already by you? I don't know PoEditor, so it won't be easy for me to see if it fits its requirements.

jbalsas:
Hey @ipeychev, I'm currently on vacation, so no, I haven't reviewed it.

The only requirement is to have the strings to be translated in a json
file. We also aimed to simplify this a little bit.

I can review next week after I get back.

ipeychev:
Ah, okay then. Great, let's manage this PR the next week.

Thanks,

jbalsas:

jbalsas:
Hey @ipeychev, I've reviewed, merged and pushed this to keep exploring the integration. In essence, this can be summarized as:
- We have a `src/ui/react/src/assets/lang/vendor/ckeditor.json` file with the strings we want to grab from CKEditor
- We have `src/ui/react/src/assets/lang/{locale}.json` files with our custom strings translated to the desired language
- We have `lib/lang/{locale.js}` files with all the CKEditor lang strings out of the box

When building we hash those three sets and if it doesn't match the previous one, we proceed to update the generated language files by simply merging the translated keys in `ckeditor.json` and the corresponding `{locale}.json`.

We've also identified some edge cases in the build that we'll proceed to fix during the week.

ipeychev:
Het @jbalsas, that's cool! My note and a small concern would be about the hash used there - it is SHA1, isn't it? SHA1 is obsolete and should be avoided. The library used in the code is able to generate hash using other algorithms too. You may think to change the algorithm to SHA226 or even use Node [crypto](https://nodejs.org/api/crypto.html) directly.

Thanks,

jbalsas:
Thanks for that tip, @ipeychev, we'll take a look to those!

PR #706, opened 2017-03-22
Remove template.json. Setup crowdin

antoniopol06:


jbalsas:
We'll do this in just one step

PR #718, opened 2017-04-05
Plugin discovery Feature

jbalsas:
Hey guys!

This is most of the work needed for the [Button Plugin Discovery Feature](https://github.com/liferay/alloy-editor/issues/667) that we've been wanting to implement for some time now.

Currently, this is just missing some API design and love, so I come to you for counsel ☺️ 

The main problem right now is that the button resolution is asynchronous, so code like this doesn't produce any buttons because those aren't resolved at the time of the invocation.

```javascript
var editor2 = AlloyEditor.editable('editable', {
    extraPlugins: AlloyEditor.Core.ATTRS.extraPlugins.value + ',ae_uibridge,ae_buttonbridge',
        toolbars: {
            styles: {
                selections: [
                    {
                        name: 'text',
                        buttons: AlloyEditor.getButtonsFrom('basicstyles'),
                        test: AlloyEditor.SelectionTest.text
                    }
                ]
            }
        }
    });

setTimeout(function() {
    console.log(AlloyEditor.getButtonsFrom('basicstyles'));
}, 1000);
```

I've added a proposal for this, so we can do something like:

```javascript
buttons: function() {
    return AlloyEditor.getButtonsFrom('basicstyles');
}
```

What do you guys think?


ipeychev:

ipeychev:
Hey Chema,

About the async button resolution, I think this should be fine. Creating the instance of the editor was always asynchronous, so for the user there shouldn't be any difference at the end.

About the code, it is fine too. Some minor things, I will comment there.

As whole, this has +1 from me :)

src/core/plugins.js
```diff
@@ -0,0 +1,66 @@
...
+     */
+    CKEDITOR.plugins.load = CKEDITOR.tools.override(CKEDITOR.plugins.load, function(pluginsLoad){
+        // Wrap original load function so we can transform the plugin input parameter
+        // before passing it down to the original callback
+        return function(names, callback, scope) {
+            pluginsLoad.call(this, names, function(plugins) {
+                if (callback) {
+                    // Iterate over plugins in the same way the callback will do, for consistency,
+                    // and wrap all plugins lifecycles
+                    for (var pluginName in plugins) {
```
ipeychev: We could use Object.keys, otherwise we have to wrap with `Object.prototype.hasOwnProperty.call(plugins, pluginName)`

src/core/plugins.js
```diff
@@ -0,0 +1,66 @@
...
+                        phase: methodName,
+                        plugin: plugin
+                    };
+
+                    return function(editor) {
+                        editor.__processingPlugin__ = payload;
+
+                        originalPluginMethod.call(this, editor);
+
+                        delete editor.__processingPlugin__;
```
ipeychev: `delete` is better to be still avoided, haven't heard engine developers resolved the performance penalties of using it. Would it be possible to use `editor.__processingPlugin__ = null` here?

jbalsas: Yeah, that's what I originally wrote, but I expected we could use this by now so we didn't leave anything behind... I'll update it :)

src/core/plugins.js
jbalsas: I was replicating https://github.com/ckeditor/ckeditor-dev/blob/master/core/editor.js#L534-L539 to make sure we iterate over the same things... what do you think?

jbalsas:
Hey @ipeychev, about the API, I was thinking we could have two APIs like:

```javascript
// Sync return based on bridged plugins
AlloyEditor.getButtonsFrom('basicstyles'); // return ['Bold', 'Italic', 'Underline'];

// Sync return will return back an array with the pluginname if no plugin is found instead of an empty array
AlloyEditor.getButtonsFrom('foo'); // return ['foo'];
```

```javascript
// Returns an async function
AlloyEditor.getButtons = function(buttons) {
    return function(editor) {
        buttons.reduce(function(acc, val) {
            val = AlloyEditor.getButtonsFrom('val');
            return acc.concat(val);
        }, []);

        return buttons;
    };
};
```

Then, this could be use like:

```javascript
var editor2 = AlloyEditor.editable('editable', {
    extraPlugins: AlloyEditor.Core.ATTRS.extraPlugins.value + ',ae_uibridge,ae_buttonbridge',
        toolbars: {
            styles: {
                selections: [
                    {
                        name: 'text',
                        // Should add the buttons ['Bold', 'Italic', 'Underline', 'link', 'twitter']
                        buttons: AlloyEditor.getButtons(['basicstyles', 'link', 'twitter']),
                        test: AlloyEditor.SelectionTest.text
                    }
                ]
            }
        }
    }
);
```

What do you think?

src/core/plugins.js
ipeychev: I see. I'm sure they haven't realized that code like those could be fragile, especially when you drop it in a site you don't control. Object.keys is a much better alternative.

jbalsas: Okay, changing it, then :)

PR #723, opened 2017-04-07
Add test to check that appendProtocol is added or not after editing Link

antoniopol06:


jbalsas:

src/ui/react/test/button-link-edit.jsx
```diff
@@ -16,56 +16,6 @@
 
         afterEach(Utils.afterEach);
 
-        it('should create link with default http protocol', function() {
```
jbalsas: Hey @antoniopol06, why are we removing these tests?

test/core/test/link.js
```diff
@@ -6,399 +6,487 @@
     describe('Link', function() {
         this.timeout(35000);
 
-        before(Utils.createCKEditor);
+        describe('general behaviour', function() {
```
jbalsas: This whole change is impossible to track down. I've been trying to put it in order hoping to see just some additions, but it turns out that several tests are now missing without explanation and others are changing assertions.

Please, we need to redo this. The default suite should stay mostly the same albeit specific needed changes that should be properly explained. Then, we can add an additional `describe('when appendProtocol is false')` to test the additional case.

src/core/link.js
```diff
@@ -160,15 +160,20 @@
          * @param {Object} modifySelection A config object with an advance attribute to indicate if the selection should be moved after the link creation.
          */
         update: function(attrs, link, modifySelection) {
+            var instance =  this;
+
             link = link || this.getFromSelection();
 
             if (typeof attrs === 'string') {
+                var href = instance._getCompleteURI(attrs);
```
jbalsas: For consistency, call this `uri`?

src/core/link.js
```diff
@@ -160,15 +160,20 @@
          * @param {Object} modifySelection A config object with an advance attribute to indicate if the selection should be moved after the link creation.
          */
         update: function(attrs, link, modifySelection) {
+            var instance =  this;
```
jbalsas: Extra spaces

PR #728, opened 2017-04-11
Updates test message and adds default case

jbalsas:
Hey @ipeychev, I've rebased and fixed the conflicts. Turns out the test file was conflicting with some previously added ones... I think it's the first time we run into problems by writing too many tests 😂

### DIFF
--- a/src/ui/react/src/adapter/alloy-editor.js
+++ b/src/ui/react/src/adapter/alloy-editor.js
@@ -43,12 +43,12 @@
 
             editor.config.selectionKeystrokes = this.get('selectionKeystrokes');
 
+            AlloyEditor.Lang.mix(editor.config, config);
+
             if (CKEDITOR.env.ie) {
                 editor.config.extraPlugins = editor.config.extraPlugins.replace('ae_dragresize', 'ae_dragresize_ie');
                 editor.config.removePlugins = editor.config.removePlugins.replace('ae_dragresize', 'ae_dragresize_ie');
             }
-
-            AlloyEditor.Lang.mix(editor.config, config);
 
             editor.once('contentDom', function() {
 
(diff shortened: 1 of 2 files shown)


ipeychev:

ipeychev:
Umm, let me see if I understood correctly the issue:
If the **user** passes explicitly `'ae_dragresize` in extraPlugins property, then the drag in IE11 will stop working. Is that correct?
```
var content = new AlloyEditor.editable('myContent', {extraPlugins: 'ae_dragresize'})
```

And why do we do [this](https://github.com/liferay/alloy-editor/pull/728/commits/e6c4abfa40e5d19ad0a8e890850bac131816b43f), user's preferences **always** have precedence?

jbalsas:
Hey Iliyan!

Yeah, on top of user preferences, I think we always want our swap to prevail. We added that so users would only need to add `ae_dragresize` (always) and we'd take care of using the right plugin depending on the browser. Also, in some cases like server-side generated config, they won't even be able to decide which one should they use. That's why we always need to do the replace at the end.

```javascript
if (CKEDITOR.env.ie) {
    editor.config.extraPlugins = editor.config.extraPlugins.replace('ae_dragresize', 'ae_dragresize_ie');
    editor.config.removePlugins = editor.config.removePlugins.replace('ae_dragresize', 'ae_dragresize_ie');
}
```

ipeychev:
Thank you, pull request merged! [See changes here.](https://github.com/liferay/alloy-editor/compare/866ebac6~1...866ebac6) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

ipeychev:
Okay... then we can merge this one for now. Thanks!

PR #729, opened 2017-04-18
Added tests to cover link when appendProtocol is false

antoniopol06:


jbalsas:
Hey @antoniopol06, [this](https://github.com/liferay/alloy-editor/pull/729/commits/466aefac048c9a24e9117c38a0bbefedbe671950) got me thinking... and I seem to remember that we did this on purpose, to allow the user to delete the protocol if he wished.

That was before we had the `appendProtocol` feature in place, so it might make sense to change this now, but I'm not 100% sure.

@ipeychev, what do you think?

ipeychev:

ipeychev:
Hey @jbalsas,

The whole flow was intentional. On creating the link, we were adding the protocol, but in case of editing, we intentionally were leaving it as entered by the user. This was a heuristic approach, of course; we tried to guess what would be useful for the user and wanted to give him the last choice how the link should look like.

Since there is `appendProtocol` property however, I would say for consistence we have to respect it on editing too.

If I have to design it now, I would probably name it not `appendProtocol`, but `linkProtocol` (or simply `protocol` if it was part of `link` configuration and not a global one) and it would be not a boolean, but a string. The default value would be '' or null. Then, the user would be able to set 'http://', 'https://', '/' or something else.

In turn, the developer would be able to set the value to 'http', 'https' or '/' and not to `true`, as it is now, because this `true` doesn't say what would be the protocol of the link after all. However, since this is a part of the public API, we either have to add another property `linkProtocol` and deprecate the current one and later remove it, or to leave it and apply the protocol when we edit the link, as it is in this PR.

Thanks,

jbalsas:
Hey @ipeychev, thanks a lot for the analysis, I totally agree with you!

I'll review and merge this as it is, and then we'll open a feature request to implement the `linkProtocol` API you drafted and deprecate this one.

Thanks1

test/core/test/link.js
```diff
@@ -400,5 +400,96 @@
             assert.isTrue(range.endContainer.equals(range.endContainer));
             assert.strictEqual(range.endOffset, 1);
         });
+
+        describe('when appendProtocol is false', function() {
+
+            it('should not add default protocol when creating a link', function() {
+                if (CKEDITOR.env.ie) {
```
jbalsas: Hey @antoniopol06, just out of curiosity... is this test really failing in IE? Is the feature working? Same for the others below...

antoniopol06:
Hey @jbalsas ,
You are right, I removed these lines of code from this PR.

jbalsas:
merged, thanks!! :)

PR #734, opened 2017-05-03
Update generated language files (auto-generated) Fixes #733

### LINKED ISSUE #733: Cannot read property 'replace' of undefined when Language is en-gb
This error occurs because of #702 does not take any language by default. Then AlloyEditor has some own keys to translate it and if object, in this case [en-gb.json](https://github.com/liferay/alloy-editor/blob/master/src/ui/react/src/assets/lang/en-gb.json) is empty. AlloyEditor.Strings will not have these strings and CKEDITOR.replace will fail.

antoniopol06:


jbalsas:
Merged!

Issue #737, opened 2017-05-10
ColorButton with Alloy Editor: Cannot read property 'focus' of undefined


jaffnae:
I'm trying to use the colorbutton plugin. I have added all the dependencies as [described in the documentation](http://alloyeditor.com/guides/ckeditor_plugins/ui_panel_menu_button/). I can see the Text Colour icon show up in the tool bar and am able to edit the font colours. However as soon as I edit the color, an error gets thrown in my console: "Uncaught TypeError: Cannot read property 'focus' of undefined"

The error is quite strange because I basically minimised the code to a stand-alone page which contains no additional code other than what's been specified on the page I linked above. I've also ensured I have all the dependencies in the Plugin folder (floatpanel, panel, button, panelbutton and colorbutton).

Here's the stand-alone page that I'm hosting specifically on Heroku. You can highlight a word and change it's color (and then observe the JS error getting thrown in the inspector): https://tempapp53.herokuapp.com/

jbalsas:
Thanks a lot @jaffnae, that's extremely useful!!

I've scheduled this issue for our next milestone and will be taking a look shortly.

jaffnae:
Thanks for  that. Can I ask if you have any timelines as to when you'll have more info on this?

PR #745, opened 2017-05-17
Panel button CKEDITOR plugin registers and creates buttons via:      1 - editor.ui.addPanelMenuButton(name, definition)      2 - editor.ui.add(name, CKEDITOR.UI_PANELBUTTON, definition)

antoniopol06:


jbalsas:
Hey @antoniopol06, there is an extra commit in this PR, about #737. How is it related to this PR? Can you send them separatedly?

Also, we are moving the registration point... this should be done consistently for all of our bridge plugins even if we haven't been able to find an instance of that usage in the wild. If need be, this could be moved to the `uibridge` plugin.

Can you please look into that and resend?

Thanks!

PR #772, opened 2017-09-28
update version of require-dir to work with Node 0.10+

vbence86:
### Intent
To update `require-dir` module so that it works well with Node 0.10.6 or above.

### Reference
`require.extensions` has been deprecated since node v0.10.6 and upgrading to node 8, which fully drops support for it, causes that error.

https://stackoverflow.com/questions/44298322/node-8-0-0-and-npm-4-2-0-error-express-load-require-extensions-hasownproperty-is

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "lodash": "4.17.4",
     "mkdirp": "0.5.1",
     "mocha": "3.2.0",
-    "require-dir": "0.3.0",
+    "require-dir": "0.3.2",
     "run-sequence": "1.1.2",
     "sinon": "1.15.4",
     "walk": "2.3.9",


jbalsas:
This was pushed directly by @ipeychev in https://github.com/liferay/alloy-editor/commit/e636edfba55c197bd9d9b693af13b50b443fde3b

ipeychev:
Hey @vbence86,

Sorry for not taking your PR but fixing it by myself. That was not ignorance.
This morning after DevCon I went over the issues and the [first](https://github.com/liferay/alloy-editor/issues/773) one I saw was "Build problem" and an exception when building AlloyEditor. Of course, this had to be immediately fixed.
The PR you sent some time ago I saw, but this time we had different priorities and it looked more like an improvement (with low priority) rather than a PR which fixes the build process. The title was "update version of require-dir to work with Node 0.10" - we don't care about Node 0.10 and the PR was not associated with an issue where it was stated that build breaks on the current version of Node.
Next time you have to add more drama there :D

Thanks,

vbence86:
@ipeychev Absolutely, I fully understand the situation. I will try to be more specific next time, though I wrote 0.10+ which I meant all versions above 0.10, but then again, I see your point.

PR #780, opened 2017-10-19
remove 3rd party react components

jbalsas:


jbalsas:
Hey @ipeychev, could you double-check this before I merge?

We've done here the bare-minimum, keeping `v0.14.2` as our bundled version of React and just applying the necessary changes so that the `*-no-react` version of `AlloyEditor` works when overlayed on top of a React16 version.

With this, we avoid any breakage or unexpected behaviour changes. This is in a big part driven by our current build process which makes adopting newer versions of React a bit more cumbersome than expected. We plan to address this soon after we release `1.5.0` with an in-depth refactor of how modules are written and bundled so we can release a `2.0.0` bundling React16.

Please, let us know what you think. Hope that makes sense!

ipeychev:
👍 from me and congratulations to @vbence86 for the great job on that!

Issue #766, opened 2017-08-28
Image Center Alignment


jaffnae:
Whenever I center-align an image, the following CSS is added:
`margin-left: 50%;
transform: translateX(-50%);`

I'd like to instead center an image in a similar way that **paragraphCenter** works. Is it possible to customize or change the behaviour of **imageCenter** button?

I've tried adding paragraphCenter into the toolbars but it still adds the above style onto the image. 

(Background info: the HTML generated inside Alloy Editor can end up inside an HTML email and most email clients do not pick up the _transform: translateX(-50%);_ CSS correctly).

vbence86:
As the preliminary station of moving towards a more flexible way to align elements, we've first updated the default code for center-alignment. With that, the `transform` attribute is no longer employed and we solely use `margin-left` and `margin-right` to achieve the same behaviour. 

We've tested it against the major email rendering engines and it looks like a more standard and reliable solution since the majority of them display the image properly.

For further information please check out the Pull Request https://github.com/liferay/alloy-editor/pull/788. 

Looking forward, we are aiming to provide more freedom for developers to specify how command are executed and how they alter the selected markup.

wincent:
By way of "spring cleaning", I'm closing issues tagged with "question" that haven't had any activity in > 12 months. There have been quite a few releases since then and much has changed, and these "stale" issues are extremely unlikely to be acted on. Please feel free to comment and we can re-open on a case-by-case basis.

PR #788, opened 2017-11-02
fix for Image Center Aligment issue

vbence86:
## Disclaimer
Use of `transform` attribute is not broadly supported by HTML Email preprocessors resulting in unexpected display of HTML constructed by **AlloyEditor**. 

For more informations please check #766.

## Intent
To provide an implementation supported by HTML Email processors to align images to the centre of the area which incorporates it.

## Solution
-  After a good deal of research about what the best way is to center-align images in HTML Emails, I removed `transform` attribute and utilized `margin-left:auto` along with `margin-right:auto` attributes.

ezgif com-video-to-gif 2

#### Tested against the following HTML Email rendering engines
- Outlook 2003
- Outlook 2008
- Thunderbird 2
- Thunderbird 3
- Gmail
- iPhone 6 iOS 8
- Apple Mail
- Gmail on Android 5.1.1

captura de pantalla 2017-11-02 a las 13 36 31

captura de pantalla 2017-11-02 a las 14 14 55

captura de pantalla 2017-11-02 a las 14 16 49


jbalsas:
Hey @vbence86, this looks quite impressive, nice work!

Just a couple of thoughts:
- Last fix seems to have broken some tests
- Our usual approach is to provide support by enhancing the configurability and extensibility of our products and sensible defaults.

Rather than trying to reach full email support (which we never intended), we should make the image alignment commands configurable so users with that need can fix that on their own.

I think the first change makes sense as a sensible default, so I would propose we merge just that and then, if needed, work on ways to make the plugin configurable.

What do you think?

vbence86:
I agree on the points you've brought up. Besides, HTML Email renderers should actually keep their tech aligned with modern standards. My venture seemed only reasonable due the small effort it needed to transform the default behavior into a sort of de-facto cross-platform standard. But with the tests failing with no apparent reason, I'm happy to step back.

Issue #768, opened 2017-09-18
Place-holders not showing anymore (after upgrading from v1.2.5 to v1.4.1)


Tuubz:

Tuubz:
add "ae-placeholder" to class

jbalsas:
Hey @Tuubz, we'll take a look at this for our next milestone. Could you add some additional steps to know how to test for it? It's hard to tell just from the description.

Thanks!

vbence86:
The fix is implemented and the related Pull Request created - https://github.com/liferay/alloy-editor/pull/789.

vbence86:
#### Steps to reproduce 
1. Add content and transforms it into `<h1>` or `<h2>` element. 
2. Empty the content of the header element. 
3. Placeholder message does NOT appear anymore since these tags are not deemed as elements that can potentially be empty.

wincent:
Doing some "spring cleaning" by closing issues that were opened in 2017 — much has changed since then and there have been many releases, so there is a risk that these older issues have grown very stale and are very unlikely to be acted upon. Please feel free to commend and we can look at re-opening on a case-by-case basis.

PR #789, opened 2017-11-06
Issue/768

vbence86:
## Prelude
Empty `<p>` elements are regarded empty content that causes the editor to show the placeholder message (`data-placeholder="Write some content here..."`). During certain scenarios the placeholder message fail to appear. 

## Issue
Check #768.

#### Steps to reproduce #768 
1. Add content and transforms it into `<h1>` or `<h2>` element. 
2. Empty the content of the header element. 
3. Placeholder message does NOT appear anymore since these tags are not deemed as elements that can potentially be empty. 

## Solution
- Extended the list of expected empty values in the `ae-placeholder` plugin. 

ezgif com-video-to-gif 4

### DIFF
--- a/src/plugins/placeholder.js
+++ b/src/plugins/placeholder.js
@@ -15,7 +15,11 @@
     var brFiller = CKEDITOR.env.needsBrFiller ? '<br>' : '';
 
     var enterModeEmptyValue = {
-        1: ['<p>' + brFiller + '</p>'],
+        1: [
+            '<p>' + brFiller + '</p>',
+            '<h1>' + brFiller + '</h1>',
+            '<h2>' + brFiller + '</h2>'
+        ],
         2: ['', ' ', brFiller],
         3: ['<div>' + brFiller + '</div>']
     };
--- a/test/plugins/test/placeholder.js
+++ b/test/plugins/test/placeholder.js
@@ -11,9 +11,7 @@
             '<h2>' + brFiller + '</h2>'
         ],
         2: ['', ' ', brFiller],
-        3: ['<div>' + brFiller + '</div>'],
-        4: ['<h1>' + brFiller + '</h1>'],
-        5: ['<h2>' + brFiller + '</h2>'],
+        3: ['<div>' + brFiller + '</div>']
     };
 
     var placeholderClass = 'ae-placeholder';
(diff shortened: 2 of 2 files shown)


src/plugins/placeholder.js
```diff
@@ -15,7 +15,11 @@
     var brFiller = CKEDITOR.env.needsBrFiller ? '<br>' : '';
 
     var enterModeEmptyValue = {
-        1: ['<p>' + brFiller + '</p>'],
+        1: [
+            '<p>' + brFiller + '</p>',
+            '<h1>' + brFiller + '</h1>',
+            '<h2>' + brFiller + '</h2>'
```
jbalsas: Hey @vbence86, I'd bet this is not enough... what about `<h3>`, or `<h4>` or ...

This bit of code has a lot of history. IIRC, this was the most robust we could come up with after a lot of trial and error. We used to just check if `editor.getData() === ''` or something like that.

Maybe this could be generalized?

Actually, from your steps, I can think of a different case that might fail: Apply a transformation that just adds attributes to the `p` or any other tag, so at the end, the empty content looks like `<p data-foo="bar"></p>`.

vbence86: Absolutely, I had all intention to make this way more generic than it is now, added a piece of code that looped through all the elements in the editor, excluded `<br>` elements along with all other whitespace characters and determined whether the editor's content is empty or not. Despite having worked pretty good, few of the tests cases began to fail due to unexpected appearance of the `data-placeholder` text. At that point I understood that few of the content tags that can be placed into the editor had a visual representation regardless the content of them. For example.: `<code>`, `<pre>`, `` and so on. As you pointed out there are more than just a few scenarios when the current approach fails. 
```
<p></p>
<p><br><br></p>
<p><br><br><br></p>
<h1></h1>
<h1><br><br></h1>
<h2></h2>
<h2><br><br></h2>
```

Also should we regard multiple `<br>` (e.g.: `<p><br><br><br><br></p>`) as an intended way to add spacing between elements and therefore make the placeholder disappear? 

There are lot clarify if we want this to be more universal. I'll open an issue for refactoring this logic and create a pull request with my intial regex solution. In the meantime, in order to ensure velocity for 1.5.0 we could carry on with the extended version of the current solution.

And lastly as for other tags like `<h3>`, I haven't found any way to reproduce this issue with `<h3>`, `<h4>` etc. headers. These headers cannot be selected through the inline toolbox.

captura de pantalla 2017-11-07 a las 10 59 31

jbalsas:
Hey @vbence86, should we simply close this, then, in favour of #795?

Issue #793, opened 2017-11-07
Add Issue template to the repository


vbence86:
## User Story  
As a developer,
I'd like to enforce guidelines how the issues are added to the repository and improve readability of what changes the assigner might apply against the codebase
so that picking up and addressing issues could take less effort for contributers.

## Acceptance Criteria
- Issue template is composed and added to `ISSUE_TEMPLATE.md` file.

PR #794, opened 2017-11-07
add ISSUE_TEMPLATE for contributers on GitHub

vbence86:
## Prelude
Issues are the main channels to capture required or desired changes to the project and therefore must be as verbose as possible. To provide an thorough understanding on what is expected to be changed in the scope of the issue we enforce essential informations to be available in order to the issue could be assigned and addressed.

## Issue
#793

## Solution
Following the guidelines of GitHub, I've added `ISSUE_TEMPLATE.md` file to the repository.

## Test
None


jbalsas:
@zenorocha, could you weigh in on this as well?

jbalsas:
@Robert-Frampton, can you take a look as well, so we can come up with something that we can apply to every other project?

Issue #792, opened 2017-11-07
Refactor code to determine when to display placeholder text


vbence86:
## Prelude
Fixing issue/768 raised concerns whether there might be a more consolidated solution to determine when the editable content is regarded empty and when the predefined placeholder text must be shown. 

For further information: 
https://github.com/liferay/alloy-editor/pull/789

## User story
As a developer,
I'd like to refactor the logic that determines whether the editable content is empty and the placeholder text should be displayed
so that we can eliminate potential technical debt building up around this. 

## Acceptance criteria
- the logic is flexible to various combinations of non visual elements
- as opposed to how it is implemented now, the logic doesn't contain any hard-coded bits with respect to DOM elements such as `<p>` or `<h1>`.
- All related test cases are aligned to the refactored code
- All tests pass
- when the editable content takes in only non visual elements the `data-placeholder` attribute must be shown. 
- non visual elements are `<br>` and inline elements with no content such as `<span>`, `<p>`, `<strong>`, `<b>`, `<i>`, `<h1>`, `<h2>`, `<h3>`, `<h4>`, `<h5>`.

## BDD scenarios
GIVEN the editable content is unfocused
WHEN it incorporates only whitespace characters or non visual DOM elements
THEN the predefined placeholder text in `data-placeholder` attribute must be shown.

wincent:
Doing some "spring cleaning" by closing issues that were opened in 2017 — much has changed since then and there have been many releases, so there is a risk that these older issues have grown very stale and are very unlikely to be acted upon. Please feel free to commend and we can look at re-opening on a case-by-case basis.

Issue #654, opened 2016-12-11
Keyup events slow with large documents


waltonseymour:
Hi,

When using a sample text of Hamlet ([link](http://courses.cs.washington.edu/courses/cse341/07wi/handouts/hamlet.txt)), each keystroke takes around 100ms, which is considerable slower than CKEditor.

I'm working with minified builds and running in production mode. My guess is that is has something to do with [this event listener](https://github.com/liferay/alloy-editor/blob/master/src/core/uicore.js#L147).

![screen shot 2016-12-11 at 1 33 09 pm](https://cloud.githubusercontent.com/assets/7594493/21082294/1e1ccbd2-bfa7-11e6-995e-259718f7a806.png)

Any suggestions for how to reduce this latency?

waltonseymour:
Interestingly, I don't notice any latency issues on the demo on http://alloyeditor.com

ipeychev:
Hello,

I added this text to our internal [demo page ](https://github.com/liferay/alloy-editor/blob/master/src/ui/react/demo/index.html) and checked. It worked fine.
Therefore, it must be something specific for the environment where you noticed the issue. What is it? What is the browser? Have you added any additional plugins?

Thanks,

waltonseymour:
I'm using chromium and I haven't added any plugins. I can take a look at it later tonight and try to debug it a bit more.

Thanks.

waltonseymour:
After digging a bit, I've found a reproduction.

Simply pasting it in will result in raw text wrapped in a <pre> tag. If you convert the text to "normal", where each paragraph becomes a <p> tag, you'll notice the latency.

I've gotten this behavior on http://alloyeditor.com/ as well

ipeychev:
Fixed, please see the commit message for more information.

Thanks for the report.

waltonseymour:
Awesome, thanks

PR #795, opened 2017-11-07
add recursive implementation of checking whether the editable area is empty

vbence86:
## Prelude
#768 was addressed earlier in #789, but the solution covers only a certain percentage of all eventualities. A more comprehensive approach is required to avoid building up technical debt. This PR aims to introduce a new approach of checking whether a given element contains content that can be regarded as empty.

## Issue
Check #792.

#### Steps to reproduce #768 
1. Add content and transforms it into `<h1>` or `<h2>` element. 
2. Empty the content of the header element. 
3. Placeholder message does NOT appear anymore since these tags are not deemed as elements that can potentially be empty. 

## Solution
- Added recursive logic that checks through the content against non-visual elements and determines whether they are empty or not. 

ezgif com-video-to-gif 4

### DIFF
--- a/src/plugins/placeholder.js
+++ b/src/plugins/placeholder.js
@@ -6,6 +6,18 @@
     }
 
     /**
+     * List of visual elements that must cause the placeholder text to disapper
+     * @property
+     * @type {array}
+     */
+    var noPlaceHolderElements = [
+        'img',
+        'hr',
+        'pre',
+        'table'
+    ];
+
+    /**
      * CKEDITOR enterMode config set the behavior of paragraphs
      * When the content is empty CKEDITOR keeps the enterMode string
      * into the content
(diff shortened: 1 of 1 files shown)


jbalsas:
Hey @vbence86, a major concern here would be performance. Since the placeholder runs (debounced) on a change event (or keyup, I don't remember correctly without taking a look), the recursive check might slow down the whole thing.

See https://github.com/liferay/alloy-editor/issues/654 for more background.

Can you double-check to see if this implementation is a cause of concern? Might be that a whitelist-oriented solution like #768 is not as bad.

We can also leave this out of the `1.5.0` milestone so we can release soon without rushing a solution. What do you think?

jbalsas:
I'm closing this for now so we can rethink it with calm in the future.

PR #815, opened 2018-04-13
Adding webpack 4 in the build process, upgrading to React 16.3.1 and removing support for IE 9 and 10

matuzalemsteles:
## New
* New build process with Webpack 4
* Update React to v16.3.1
* Use of the new refs API
* Install `transform-class-properties` for` static` use in class

## Changes
* Refactor in codebase bdbd6b0
* Corrections of the tests
* Replace `React.addons.TestUtils` with `react-dom/test-utils`
* Removes the use of PropTypes

## Breakings
* Our components are no longer exposed as before by accessing the `AlloyEditor.ButtonBold` will cause `undefined` for example. We continue with `AlloyEditor.Buttons[...key]`.
* Removing support for IE 9 and 10

### Components
Made minor changes to some components, see ff816fe, f66fda1, 3d7633e...

### Tests
Many of the tests have been set to work again with the new build process, some changes consist of `imports` of files or mocks of functions.

#### Skip some test
Ref: 4ecd00c

It was necessary to pass some tests since the UIBRIDGE components have not yet been refactored and are having problems with findAllInRenderedTree and isCompositeComponentWithType.

### Use gulp for generate files
For a moment we are maintaining the use of gulp to generate the css files, icon fonts, copy files from CKEDITOR, languages and generate the demo. In the near future we will be able to remove the gulp by scripts next to the Webpack.

### CI
Test run times averaged **4 min** compared to **7 min**. 🙂

> Estimate

### Bundle
With the new build process with webpack 4, the size of some of our packages have remained the same and others have increased because of things the webpack ends up adding to create the wrapper or adding polyfill if necessary.

## New structure
* **`dist/`** - Build Files
* **`lib/`** - CKEditor Files
* **`scripts/`** - Build-related scripts and utilities for testing.
* **`src/`** - All code base, Core, Plugins and UI
* **`test/`** - All of our tests together
* **`package.json`** - npm module manifest

### Future
Our structure still needs refinements, but we need to work on our code base to continue polishing architecture over time.

* Using [Flow](https://flow.org/) in code base 🎉🙂
* Remove the use of gulp
* Remove use of `create-react-class`
* Remove use of the [lifecycle APIs from the React](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html) `componentWillMount`, `componentWillReceiveProps` and `componentWillUpdate`.


jbalsas:
This looks like an awesome starting point, @matuzalemsteles!

I'm merging this in so we can continue on top of it. 

There are some concerns regarding [CKEditor License change](https://github.com/ckeditor/ckeditor5/issues/991), so we might need to stick with `4.x` for our `2.x` version...

jbalsas:
I've merged this into `develop` and forked `1.x` branch for future maintenance of the existing version

PR #837, opened 2018-05-11
AUI-3148

jonmak08:
Hey @liferay,

Attached is an update for http://issues.liferay.com/browse/AUI-3148.

> ## Prelude
> A customer noted to us that users are unable to drag and drop text or images internally in AlloyEditor. I verified that this is not only the case in the portal but on alloyeditor.com as well.
> 
> ## Issue
> https://issues.liferay.com/browse/AUI-3148
> 
> ## Solution
> I made it so the preventDefault is only fired by AlloyEditor's dragdrop function when we need to specifically override the behavior, otherwise the default CKEditor dragDrop should be use.
> 
> ## Test
> Tested on Chrome, Firefox, MS Edge, and IE11.

Please let me know if you have any questions. Thanks!

### DIFF
--- a/src/plugins/addimages.js
+++ b/src/plugins/addimages.js
@@ -125,17 +125,15 @@
 
                 var transferFiles = nativeEvent.dataTransfer.files;
 
-                if (transferFiles.length <= 0) {
-                    return;
+                if (transferFiles.length > 0) {
+                    new CKEDITOR.dom.event(nativeEvent).preventDefault();
+
+                    var editor = event.listenerData.editor;
+
+                    event.listenerData.editor.createSelectionFromPoint(nativeEvent.clientX, nativeEvent.clientY);
+
+                    this._handleFiles(transferFiles, editor);
                 }
-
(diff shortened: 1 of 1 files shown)


jbalsas:
Hey @jonmak08, @gregory-bretall, could you please create an issue in this repo and remove the `AUI-` reference in the commits? As it can be reproduced in the standalone AlloyEditor distribution, it's better to avoid mixing projects and having to jump from one place to another to find the issue.

I'm keeping this open and I'll do this myself on wednesday if you guys don't have time for it.

jbalsas:
Oh, I see. Here we should just refer to the issue in this repo. The commit messages should read as:

```
Do not prevent default drop behavior unless a file is being dropped from outside of the editor.

Fixes #839
```

As this has nothing to do with AlloyUI

gregory-bretall:
Thanks @jbalsas I resent it with the messages changed to the AlloyEditor repo, PR is here https://github.com/liferay/alloy-editor/pull/840

Issue #848, opened 2018-06-01
Toolbar doesn't close after hyperlinking an image


SpencerWoo:
## Prelude
In [#128](https://github.com/liferay/alloy-editor/issues/128 ) we intentionally advance the selection and intend to close the toolbar on submission / action completion.

When hyperlinking images we use [nextRenderedElement](https://raw.githubusercontent.com/liferay/alloy-editor/master/dist/alloy-editor/alloy-editor-no-ckeditor.js)
`if (shouldUpdateReactComponent(prevRenderedElement, nextRenderedElement))`
to determine whether the component should be unmounted.
`ReactReconciler.unmountComponent(prevComponentInstance);`

The issue is that the toolbar does not unmount as nextRenderedElement is valid.
https://imgur.com/4qTbzFM

## Reproduce
1. Downloading [AlloyEditor 1.5.0 source](https://github.com/liferay/alloy-editor/releases/download/1.5.0/alloy-editor-1.5.0.zip)
2. Modify index.html to import the non-minified version of alloy-editor
`<script src="alloy-editor/alloy-editor-all.js"></script>`
3. Modify alloy-editor-all.js to add 'link' button to [image toolbar](https://github.com/liferay/alloy-editor/blob/master/dist/alloy-editor/alloy-editor-core.js#L7270)
`buttons: ['imageLeft', 'imageCenter', 'imageRight', 'link']`
4. Open the editor with Chrome.
5. Add an image or click the existing one
6. Add a hyperlink.  

Result : 
On submission the toolbar isn't closed.  However the toolbar closes when hyperlinking text.

## Acceptance Criteria
- Hyperlinking toolbar closes after submission.
- Test in Chrome, IE11, and FireFox.

PR #849, opened 2018-06-01
Image link dev

SpencerWoo:
## Issue
https://github.com/liferay/alloy-editor/issues/848

## Solution
For most browsers "check if selection is empty" is sufficient to resolve this issue.

However the Chrome react unmount issue is different as CKEditor injects [7 zero-width space characters].  We test whether the selection is an image in linkSelectionTest to handle this.

## Test
Chrome, IE 11, and FireFox.

gif of issue and fix on Chrome
https://imgur.com/a/h08XgJt

### DIFF
--- a/src/ui/react/src/selections/selection-test.js
+++ b/src/ui/react/src/selections/selection-test.js
@@ -30,11 +30,17 @@
     var linkSelectionTest = function(payload) {
         var nativeEditor = payload.editor.get('nativeEditor');
         var range = nativeEditor.getSelection().getRanges()[0];
+        var selectionData = payload.data.selectionData;
+
+        var selectionDataName;
 
         var element;
 
+        if (selectionData.element) selectionDataName = selectionData.element.getName();
+
         return !!(
             nativeEditor.isSelectionEmpty() &&
+            selectionDataName !== 'img' &&
             (element = (new CKEDITOR.Link(nativeEditor)).getFromSelection()) &&
             element.getText().length !== range.endOffset &&
             !element.isReadOnly() &&
--- a/src/ui/react/src/selections/selection-test.js
+++ b/src/ui/react/src/selections/selection-test.js
@@ -45,9 +45,18 @@
     var imageSelectionTest = function(payload) {
         var selectionData = payload.data.selectionData;
 
+        var selectionEmpty = false;
+
+        if (payload.editor) {
+            var nativeEditor = payload.editor._getNativeEditor();
+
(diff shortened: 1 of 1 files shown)


jbalsas:
Hey @SpencerWoo, this is actually breaking the image selection test...

```
 SelectionTest image should detect an image selection FAILED
```

SpencerWoo:
Hi @jbalsas and @antonio-ortega, I've had difficulty figuring out how to proceed with this one so I'd appreciate your input

This PR fails this test https://github.com/liferay/alloy-editor/blob/25151898f9385f09096f05ab7cf30aeea281330d/src/ui/react/test/selection-test.js#L112-L118

The test is always expecting `True` from `imageSelectionTest()` however our change is that we return `False` from `imageSelectionTest()` when `isSelectionEmpty()` is `True`.

We use the `isSelectionEmpty()` check in `imageSelectionTest()` because

- it is `False` when the image is selected (has a toolbar)
- it is `True` after the user confirms/submits the hyperlink (because range[0].collapsed becomes true)

https://github.com/liferay/alloy-editor/blob/25151898f9385f09096f05ab7cf30aeea281330d/src/core/selection-region.js#L202-L209

This `isSelectionEmpty()` check here is also in `textSelectionTest()` however those don't cause textSelection to fail tests.

<hr>

I've had trouble running the tests locally but does it make sense that `imageSelectionTest()` returns `True` when `isSelectionEmpty()` is `True`? 
If it doesn't make sense then should the test be changed to be more specific and pass this case?  
If it does make sense then any input on modifying the fix (using a different check) is appreciated and very helpful.

jbalsas:
Hey @SpencerWoo, I think this shows an issue with these tests, since we aren't properly setting the selection.

I think the correct way to go would be add inside the `getPayload` helper at https://github.com/liferay/alloy-editor/blob/master/src/ui/react/test/selection-test.js#L8 the following:

```javascript
this.editor.get('nativeEditor').getSelection().selectElement(element);
```

That's likely going to make other tests fail and/or introduce some changes, so a more localized fix could be something like:

```javascript
it('should detect an image selection', function() {
    var nativeEditor = this.editor.get('nativeEditor');

    var element = nativeEditor.element.findOne('#editable');

    this.editor.get('nativeEditor').getSelection().selectElement(element);

    var payload = getPayload.call(this, element);

    assert.isTrue(imageTest(payload));
});
```

Can you work around with this and see how it looks?

Thanks!

jbalsas:
I'll actually go ahead and push the second option so we can get a release out the door!

PR #850, opened 2018-06-01
Adds a way to correct paste behavior of an URL

diegonvs:


jbalsas:
Hey @diegonvs, this looks very cool! Could you detail a little bit what the regex is actually letting through now in comparison with before? Mind adding a test while we're at it?

Also, what's the reasoning for automatically creating a link when pasting a url? For this, it should be easy to remove the link right afterwards, otherwise we might get in the way of the users. Can you double-check that interaction?

diegonvs:
- About RegEx: I've noticed that we have a weak RegEx to test our link use cases, so I've searched a lot of discussions about regex to test URLs and I noticed that it's impossible to [cover all validations of URL](https://mathiasbynens.be/demo/url-regex) but I found a good one that fits perfectly in most  use cases.
- I'll check the interaction that you describe :)
- About tests, if all user interaction is okay, I'll start type them!

diegonvs:
I'm re-testing regex to provide a better coverage for all possible URLs use cases.

diegonvs:
I've concluded my benchmark searching a lot of RegExp and is this is the best RegExp to this use case.  I'm working on tests right now! 🔥

Issue #852, opened 2018-06-11
While resizing an image some elements change visibility


SpencerWoo:
## Prelude
When resizing an image elements (context bar, view source, plus button) are initially hidden but reappear after a while.

## Acceptance Criteria
Consistency could look like either of these options
- The UI is always hidden until the resizing action is complete
- The UI is always shown while resizing.

## Notes
The editor hides because [_setUIHidden](https://github.com/liferay/alloy-editor/blob/d34fe1b67b9a70e417aac72842ead20222131903/src/ui/react/src/components/main.jsx#L347) is meant to hide the UI when you leave the editable area however dragging doesn't fit into [these conditions](https://github.com/liferay/alloy-editor/blob/d34fe1b67b9a70e417aac72842ead20222131903/src/ui/react/src/components/main.jsx#L359-L365).  Perhaps drag resizing should make that condition true so the relevant items will always be visible?

On the other side I believe [_onEditorInteraction0](https://github.com/liferay/alloy-editor/blob/d34fe1b67b9a70e417aac72842ead20222131903/src/ui/react/src/components/main.jsx#L312-L319) is what makes the toolbar / all the relevant items appear again.

PR #855, opened 2018-06-12
Resize visible

SpencerWoo:
## Issue
https://github.com/liferay/alloy-editor/issues/852

In [_setUIHidden](https://github.com/liferay/alloy-editor/blob/d34fe1b67b9a70e417aac72842ead20222131903/src/ui/react/src/components/main.jsx#L347) we check whether the user's interaction target is within the editor / editable area.  

The issue is that dragresize dynamically generates the drag handles outside of the editable div and making `editable.contains(targetNode)` false and thus we're treating resizing as though we're outside the UI.

## Solution
We modify the [_setUIHidden check](https://github.com/liferay/alloy-editor/blob/d34fe1b67b9a70e417aac72842ead20222131903/src/ui/react/src/components/main.jsx#L359) to also test whether we're resizing an image.  We do this by checking whether the target is within the div with id "ckimgrsz".

## Test
Chrome and FireFox.

### DIFF
--- a/src/ui/react/src/components/main.jsx
+++ b/src/ui/react/src/components/main.jsx
@@ -349,6 +349,7 @@
 
             if (domNode) {
                 var editable = this.props.editor.get('nativeEditor').editable();
+                var parentNode = target.parentNode;
                 var targetNode = new CKEDITOR.dom.node(target);
 
                 if (!editable) {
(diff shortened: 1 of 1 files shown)


jbalsas:
Awesome, thanks @SpencerWoo !

Issue #866, opened 2018-06-27
Toolbar increases in size when close to right window edge


fortunatomaldonado:
## Prelude
When editing a text on the left side of the editor close to the window edge, the toolbar will always be a set size. When editing a text that is located close to the right window edge, the toolbar will be resized to allow all components to be viewed. This resizing can look off especially when trying to add a link.

## Reproduce
1. Go to demo at https://alloyeditor.com/
2. Resize the browser so that the right side of the browser is as close as possible to the editor.
3. Highlight a text and notice the toolbar size.
4. Click the link option and notice the toolbar size.
5. Repeat step 2, 3, and 4 but on the right side of the editor.

The toolbar will always be a consistent size until it reaches the right side of the browser.

## Acceptance Criteria
- The toolbar will be a consistent size no matter on the left or right of the browser.

## Notes
There is not resizing on the left since the [getWidgetXYPoint](https://github.com/liferay/alloy-editor/blob/master/src/ui/react/src/components/base/widget-position.js#L122) makes sure the widget will be visible on the top and left of the screen. The right side is resized due to CSS.
![ae](https://user-images.githubusercontent.com/31672551/42003200-1a2f27de-7a1f-11e8-9e91-5a37a2ab971d.gif)

jbalsas:
Fixed by https://github.com/liferay/alloy-editor/pull/867 and released with `1.5.11`

jbalsas:
Hey @fortunatomaldonado, you should simply need to backport https://issues.liferay.com/browse/LPS-83409

fortunatomaldonado:
Hello @jbalsas,

Thank you for the information. I was testing the fix in 7.0.x and I found it would overflow off the screen from time to time. I'm not sure if you also have seen this behavior. I tested some more and found we can subtract the document.body.offset by the offsetWidth to prevent this overflow in 7.0.x. This does not seem to be an issue in Master since the margin for the editor are further away from the browser edge. I have not yet tested this on Master. Let me know what you think about this.
Thank you.

Overflow
![overflow-toolbar](https://user-images.githubusercontent.com/31672551/42663396-30304104-85ea-11e8-93bf-6c80f229966c.gif)


Subtracting offsetWidth
![using-offsetwidth](https://user-images.githubusercontent.com/31672551/42663408-3dcbd45e-85ea-11e8-9f77-7ba559dd0af7.gif)

fortunatomaldonado:
Hello @jbalsas,

[...]

PR #868, opened 2018-07-17
Adjust toolbar placement on right side to prevent overflow

fortunatomaldonado:
## Issue
The toolbar may overflow on the right side of the screen if the user has a smaller screen or resizes the browser. This will prevent users from seeing toolbar options.

## Reproduce
Example of issue can be found here: https://github.com/liferay/alloy-editor/issues/866#issuecomment-404677783

## Solution
By checking if the placement of the `left` is going to be greater than the document's body offset subtracted by the toolbar width, we can prevent any overflow.

### DIFF
--- a/src/ui/react/src/components/base/widget-position.js
+++ b/src/ui/react/src/components/base/widget-position.js
@@ -148,8 +148,8 @@
             }
 
 
-            if (left > document.body.offsetWidth - halfWidth) {
-                left = document.body.offsetWidth - halfWidth;
+            if (left > document.body.offsetWidth - offsetWidth) {
+                left = document.body.offsetWidth - offsetWidth;
             }
 
             if (top < 0) {


jbalsas:
Hey @fortunatomaldonado, thanks for sending this over! Based on the screenshots, it looks like we need some adjustment. 

@julien is currently on vacation. He'll be back next week. If that's okay, I'd like to wait so he can take a look here.

Thanks!

fortunatomaldonado:
Hello @jbalsas,

Of course, that is not a problem. Thank you for letting me know.

fortunatomaldonado:
Hello @julien,

When you have a chance, can you take a look at this?

Thank you.

julien:

julien:
Hi @fortunatomaldonado,

Sorry for the delay; I just tested in master and everything seems working fine as far as I'm concerned

868_liferay-portal

But in the alloy-editor demo however it seems it's not working exactly the same way

868_alloy-editor

If this is an acceptable behaviour, I'm ok with the changes. Tell me what you think when you can.

Thanks

duracell80:
Hey guys thanks for working on these issues they will make a big positive improvement for users. Has there ever been a suggestion to move away from the cursor placement of this bar?

It seems a lot of issues would just go away if the bar was afixed top left of the text area and fixed in place.

fortunatomaldonado:
Hello @julien,

Thank you for the quick review. I saw the same results in the demo as well. I believe this is due to the demo's editor being surrounded by margins unlike in master and 7.0.x. If you remove the margins from the demo it seems to work as intended. I believe this will still be acceptable behavior.

Thank you again for looking into this.

julien:
Hi @fortunatomaldonado ,
I tested again without any margins applied and you're right, it's all good.
Thanks!

Issue #876, opened 2018-08-23
Image Toolbar extend past browser


fortunatomaldonado:
## Issue
The image toolbar will extend past the browser for certain image sizes and browser sizes. 

## Reproduce
1. Using the default AlloyEditor demo, remove the margins from the page.
2. Resize the moon image making it as small as possible.
3. Select the image and left/right align it.

You are able to see that the toolbar extends past the browser.
![aei](https://user-images.githubusercontent.com/31672551/44542090-b4ad7b80-a6c0-11e8-8a6c-563ee77adfaa.gif)

## Acceptance Criteria
The toolbar should adjust its position when it is going to extend past the browser.

julien:
Hi @matuzalemsteles,

This [test](https://github.com/liferay/alloy-editor/blob/master/src/ui/react/test/selection-position.jsx#L65) is failing, could you please take a look at fixing it?

Thanks

fortunatomaldonado:
Hello @julien,

I believe this test should be updated if the fix for the current issue is to be accepted. It is asking for the toolbar to be centered on an image, but with my current changes it adjust the toolbar so it will not overflow out the browser. 
Let me know if I am understanding this correctly or if I am off base.
Thank you.

julien:
Hi @fortunatomaldonado,

Yes you're understanding this correctly.

I think we should have a test that demonstrates the fix you implemented, if the previous test does not make sense anymore there's no need to keep it there.

Let me know if you need help.

Thanks!

jbalsas:
Hey!

Just a small note. We had a previous behaviour that we still want to maintain, so, before merging this in, I'd expect to see:
- A fix in the current test to assert that in the general case, the toolbar still shows centered on the image
- A new test explicitly checking the case where the toolbar will overflow and asserting that in that case it is where it's expected to be

Thanks!

fortunatomaldonado:
Hello @julien and @jbalsas,

Thank you both for the advice going forward. I am currently working on other issues but should get to this by the end of the week. I may need some help with this but we will cross that bridge when we get there.
Thank you both again!

fortunatomaldonado:
Hello @julien and @jbalsas,

I am beginning to work on this but I am not too familiar with how to run these tests or the basics of creating a test. Looking at the other tests, I have some what of an understanding of what is happening so I think creating a new test will be basing it off the others. 
Is there a way to run these tests locally and specify which test to run? Is there any resource I can look at regarding creating tests?
[...]

PR #883, opened 2018-09-25
Fix editor.on paste function and fix REGEX_URL

rolandpakai:
## Prelude
The issue was introduce by the following commit: 
"Adds a way to correct paste behavior of an URL"
https://github.com/liferay/alloy-editor/commit/32040b86150be5016a3e30ddaa78031a3c09b0c9

## Issue
Pasting text from the clipboard which contains url(s) the first url is automatically moving to the beginning of the paragraph replacing with an anchor link. The other urls stay untached. 

The issue is only reproducable when copying text from the clipboard
- which has url(s)
- and not contains new lines
- and the text is not copied from the editor itself.

When typeing text in the editor the url(s) are replacing with anchor links correctly.

## Solution
The original solution was wrong: only replaced the first url in the text and placed the anchor link to the beginning of the text.

I changed the REGEX_URL to find all the url occurrence in the text and i rewrited the paste function to replace all the urls in place with the anchor links. I also added the functionality to replace the valid emails with anchor links too.

https://issues.liferay.com/browse/LPS-85659

### DIFF
--- a/src/plugins/autolink.js
+++ b/src/plugins/autolink.js
@@ -25,8 +25,10 @@
 
     var REGEX_LAST_WORD = /[^\s]+/mg;
 
-    var REGEX_URL = /((([A - Za - z]{ 3, 9}: (?: \/\/)?)(?:[-;:&=\+\$,\w]+@)?[A-Za-z0-9.-]+|(https?\:\/\/|www.|[-;:&=\+\$,\w]+@)[A-Za-z0-9.-]+)((?:\/[\+~%\/.\w-_]*)?\??(?:[-\+=&;%@.\w_]*)#?(?:[\w]*))((.*):(\d*)\/?(.*))?)/i;
-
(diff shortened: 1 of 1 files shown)


jbalsas:
Hey @rolandpakai, this looks very interesting, thanks for contributing!

Right now, the change is causing some tests to fail... could you take a look to see if they're failing for valid reasons and their behaviour was wrong? Maybe we just need to udpate them, but it's also possible that something in the fix is introducing a breaking change.

Thanks!

rolandpakai:
Unfortunatelly i dont known how/where to check the reasons of the failed tests...
Could you please help me out?

And is it possible to add my changes to the next v1.5.14 release?

Thanks, Roland

jbalsas:
Looks like it's failing on `IE10` on `Windows`. You can see it here: https://travis-ci.org/liferay/alloy-editor/builds/432958631

We'll include this in the next release, indeed.

rolandpakai:
Thank you!
Could you please tell me when will be scheduled the next release?

Thanks, Roland!

jbalsas:
We'll release as soon as this and https://github.com/liferay/alloy-editor/issues/876 are fixed

PR #884, opened 2018-09-26
Fixes #876 | Fix toolbar position

### LINKED ISSUE #876: Image Toolbar extend past browser

julien:


jbalsas:
I think we agreed this still needed some additional work to meet expectations. Closing for now @julien!

PR #887, opened 2018-10-01
Fixes #876 | Fix toolbar position

### LINKED ISSUE #876: Image Toolbar extend past browser

julien:


jbalsas:
Hey @julien, this looks much better!! 👏 

I've found a bit of an edge case that I'd like to try to fix before we merge this in. Looks like when we re-render the toolbar in some cases it overflows the page. See the screencast when applying a text change or clicking on the link button on the right margin.

toolbar_movement

Could you try to figure out what's wrong there?

Thanks!

julien:
Hey @jbalsas,

Yes it looks like I totally missed that one.
I'll have a look and fix this.

Thanks

julien:
Pull request submitted to https://github.com/liferay/alloy-editor/pull/888

PR #904, opened 2018-11-14
Fix #903 | If editor is present, drag and drop into non-editor…

### LINKED ISSUE #903: If AlloyEditor is Present, Drag&Drop Should Work Only Within the Editor
## Prelude
The current behavior in the portal is that drag&drop of media content (such as images and videos) onto the browser will open the content in the browser. This behavior becomes problematic when you initiate a a drag&drop of the content within something such as a Web Content. All work will be lost (if the work has not been saved) if a user does not the drop the content within the editable areas of the editor as the content will be loaded in the browser.
I cannot reproduce the issue with AlloyEditor's demo, but the issue is reproducible in the portal.

## User story
As a user,
I'd like to have drag&drop only work within the editor if the editor is present
So that accidental drops of media content in non-editable areas do not trigger the default browser response to drag&drop.

## Reproduction Steps
Steps to Reproduce:
1. Make a new Basic Web Content with some edits and save
2. Choose to make more edits and then drag an image into the content
3. The image is now part of the web content and can be edited itself (moved, size changed, etc.)
4. Drag another image to the web content editor, but this time, fail to drop it in the content window (e.g., let go of the mouse button in wrong place somewhere outside the content window)
5. Browser now opens the file full screen
6. Only option is to hit back arrow in browser

Expected Results:
Drag and drop should be disabled in non-editable areas.

Actual Results:
Browser opens the file.

Let me know if you have any questions or comments.
Thank you.

brianikim:
## Issue

#903 
The current behavior in the portal is that drag&drop of media content (such as images and videos) onto the browser will open the content in the browser. This behavior becomes problematic when you initiate a a drag&drop of the content within something such as a Web Content. All work will be lost (if the work has not been saved) if a user does not the drop the content within the editable areas of the editor as the content will be loaded in the browser.
I cannot reproduce the issue with AlloyEditor's demo, but the issue is reproducible in the portal.

## Solution
The solution is very straight forward. If the media content is not currently being dropped into the editable area of the editor, we should prevent the default behavior of drag and drop.

I tested on Firefox and Chrome.

Thank you.

### DIFF
--- a/src/plugins/addimages.js
+++ b/src/plugins/addimages.js
@@ -7,6 +7,13 @@
         return;
     }
 
+    window.addEventListener('dragover', function(event){
+        event.preventDefault();
+    },false);
+    window.addEventListener('drop', function(event){
+        event.preventDefault();
+    },false);
+
     /**
      * CKEditor plugin which allows Drag&Drop of images directly into the editable area. The image will be encoded
      * as Data URI. An event `beforeImageAdd` will be fired with the list of dropped images. If any of the listeners


duracell80:
Could a test be run in Internet Explorer 11 too please?

src/plugins/addimages.js
```diff
@@ -7,6 +7,15 @@
         return;
     }
 
+    else {
```
julien: It doesn't look like this `else` statement is necesarry.

src/plugins/addimages.js
```diff
@@ -7,6 +7,15 @@
...
+        window.addEventListener("dragover",function(event){
```
julien: Could you please reformat this?

- Use single quotes for strings
- Add spaces after comas `, `
- Add a space before the opening braces/curly brackets of functions? 

I think that if you look at the rest of the code, it should be easy to figure out, but if you have a doubt you can always ask me for help.

Thanks!

julien:
@duracell80 we do test on IE11, usually this is done via saucelabs, but we also test manually for this kind of issue.

brianikim:
Hello @julien ,

I have updated this PR with the changes you have recommended.

Sincerely,
Brian Kim

jbalsas:
Hey @julien, @brianikim, see https://github.com/liferay/alloy-editor/issues/903 before moving forward with this, please.

brianikim:
Hello @jbalsas ,

Please take a look at my latest comment on #903.

PR #907, opened 2018-11-15
Fix autolink test

julien:


julien:
@jbalsas could you please have a look at this and share your thoughts?
I'm basically trying to fix the test that recently started failing in saucelabs, but since nothing had been changed for a long time in this file, I'd like to have a second opinion.

Thanks!

jbalsas:
Hey @julien, in which browser was this failing? The test is basically positioning the cursor after the link and simulating an `ENTER` press... sometimes, bender has trouble positioning the cursor, and this may change from one browser version to another...

I'd say this feels okay. We should run some manual tests on this to make sure, but I'll merge for now.

julien:
@jbalsas, 

After updating the saucelabs config, this test was failing on saucelabs for these browsers+os'es combinations:

Chrome 70.0.3538 on Windows 10
Chrome 70.0.3538 on OS X 10.11 (El Capitan)
Chrome 70.0.3538 on macOS 10.12 (Sierra)

The strange thing though, is that they're failing neither on macOS 10.13 (High Sierra) nor on Linux (I'm not sure about the distro used) for the same browser.

Even stranger, locally (even though this shouldn't be relevant) I have the exact same Chrome version on OS X 10.11 and it wasn't failing (neither before or after the change ... ) 

We also ran some manual tests with @antonio-ortega and the "rendered" HTML doesn't look exactly like what's expected in the tests, i.e. 

Instead of this:

```html
<p>link <a href="http://www.liferay.com">www.liferay.com</a></p><p>text</p>
```

What I see after typing:

```
link www.liferay.com text
```

Looks like this:

```html
<p>link <a data-cke-saved-href="http://www.liferay.com" href="http://www.liferay.com">www.liferay.com</a>&nbsp;test</p>
```

I understand that these tests are "difficult" to implement especially in a cross-browser way but I agree that we should do something about this.

Anyway thanks again for your help and feedback.

PR #912, opened 2018-11-16
Fixes #911 Use correct CSS to place drag resize handle at the corner

### LINKED ISSUE #911: AlloyEditor image resize handle position is misplaced in IE11 when image is right aligned
**Issue**

https://issues.liferay.com/browse/LPS-87458

In IE11 when changing the image's alignment to the right the drag resize handle moves from it's expected lower righthand location to the lower lefthand side. However, the handle is misplaced slightly to the right of the lefthand corner.

**Reproduce**

I was able to reproduce this behavior in Web Contents of Liferay using IE11. I cannot reproduce on the alloyeditor demo while using IE11.

1) Start Liferay.
2) Open IE11.
3) Add a web content article.
4) Insert an image and resize it.
5) Right align the image.

**Expected Result**: the handle is placed correctly on the left corner.
**Actual Results**: the handle is slightly misplaced off the corner.

Let me know if there are any questions or comments.
Thank you.

brianikim:
**Issue**

#911
In IE11 when changing the image's alignment to the right the drag resize handle moves from it's expected lower righthand location to the lower lefthand side. However, the handle is misplaced slightly to the right of the lefthand corner.

**Solution**

The CSS property **left** for cke_image_resizer_left just has to be adjusted to be -25px isntead of -5px.

The issue only appears in IE11 and have tested also in Chrome, Firefox, and Edge. The issue is only reproducible on Liferay AlloyEditor.

Let me know if there are any questions or comments.
Thank you.

### DIFF
--- a/src/plugins/dragresize_ie11.js
+++ b/src/plugins/dragresize_ie11.js
@@ -50,7 +50,7 @@
             // Bottom-left corner style of the resizer.
             '.cke_image_resizer.cke_image_resizer_left{' +
                 'right:auto;' +
-                'left:-5px;' +
+                'left:-25px;' +
                 'cursor:sw-resize;' +
             '}' +
             '.cke_widget_wrapper:hover .cke_image_resizer,' +


brianikim:
cc/ @jbalsas

julien:

PR #920, opened 2018-11-20
Change dragresize implementation on IE to modify heigh and width insi…

gregory-bretall:
…de the style attribute so as to be consistent with the standard AlloyEditor resizing behavior.

Please employ to following template as the base structure of Pull Requests for **AlloyEditor** (liferay/alloy-editor) project. Pull Requests not incorporating the sections below will be regarded as unsatisfying and rejected. 

## Prelude
Default dragresize implementation doesn't work in IE11, so the CKEditor behavior was copied over without changes.

## Issue
The issue is a mismatch in how dragresize modifies height/width for an image between IE and other browsers. In most browsers, we modify height and width inside the image's style attribute. However, IE's implementation of dragresize copies directly from how CKEditor handles this which is to use individual height and width attributes for the image. The problem here is if both are present because the image was edited in different browsers then the edits from non-IE browsers will always take precedence.

## Solution
I removed our use of the height/width attributes for dragresize_ie and replaced them with style modifications to line up with the behavior on other browsers.

## Test
This fix has been tested on Internet Explorer 11, the main environment were we use dragresize_ie.

### DIFF
--- a/src/plugins/dragresize_ie.js
+++ b/src/plugins/dragresize_ie.js
@@ -878,10 +878,9 @@
                 // Don't update attributes if less than 10.
                 // This is to prevent images to visually disappear.
                 if (newWidth >= 15 && (newHeight >= 15 || newHeight === 'auto')) {
-                    image.setAttributes({
-                        width: newWidth,
-                        height: newHeight
-                    });
+                    image.$.style.width = newWidth;
+                    image.$.style.height = newHeight;
+                    
                     updateData = true;
                 } else {
                     updateData = false;
@@ -902,10 +901,8 @@
                 resizer.removeClass('cke_image_resizing');
 
                 if (updateData) {
-                    widget.setData({
-                        height: newHeight,
-                        width: newWidth
-                    });
+                    widget.element.$.style.width = newWidth;
+                    widget.element.$.style.height = newHeight;
 
                     // Save another undo snapshot: after resizing.
                     editor.fire('saveSnapshot');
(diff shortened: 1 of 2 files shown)


julien:

julien:
Hi @gregory-bretall,

Thanks for looking at this.

I've got a doubt however, when I test this change I'm not able to resize images

drag_resize_after

But before this change it was working 

drag_resize_before

Could you please review this on your side?

Thanks

julien:
@gregory-bretall could you please send your PR against the `develop` branch please? 
Thanks

src/plugins/dragresize_ie.js
```diff
@@ -878,10 +878,9 @@
                 // Don't update attributes if less than 10.
                 // This is to prevent images to visually disappear.
                 if (newWidth >= 15 && (newHeight >= 15 || newHeight === 'auto')) {
-                    image.setAttributes({
-                        width: newWidth,
-                        height: newHeight
-                    });
+                    image.$.style.width = newWidth;
```
julien: Since we're using styles, shouldn't the value should be in pixels?

gregory-bretall: You're absolutely right @julien I had used pixels in my testing version of the fix and must've forgotten it when I transferred it over to send the pull. I'll resend it with that change on a new PR.

gregory-bretall:
Hi @julien I have sent an updated PR to https://github.com/liferay/alloy-editor/pull/923. Please let me know what you think.

julien:
Hi @gregory-bretall sure. thanks!

Issue #901, opened 2018-11-12
Image Width Is Being Set Differently Between Browsers Causing IE Resize Failing


duracell80:
Simply the resize plugin for IE is being made to stop working when edits are made in chrome. This is because Chrome sets width in the inline style whereas IE sets it as a HTML attribute, creating code such as img ... width=300 style=width:250px

Cross browser editing then causes the resizing function to stop working for Internet Explorer.

gregory-bretall:
A fix has been sent in for this here https://github.com/liferay/alloy-editor/pull/920

gregory-bretall:
Updated PR is found here https://github.com/liferay/alloy-editor/pull/923

PR #923, opened 2018-11-26
Change dragresize implementation on IE to modify heigh and width insi…

gregory-bretall:
…de the style attribute so as to be consistent with the standard AlloyEditor resizing behavior.

## Prelude
Default dragresize implementation doesn't work in IE11, so the CKEditor behavior was copied over without changes.

## Issue
https://github.com/liferay/alloy-editor/issues/901

The issue is a mismatch in how dragresize modifies height/width for an image between IE and other browsers. In most browsers, we modify height and width inside the image's style attribute. However, IE's implementation of dragresize copies directly from how CKEditor handles this which is to use individual height and width attributes for the image. The problem here is if both are present because the image was edited in different browsers then the edits from non-IE browsers will always take precedence.

## Solution
I removed our use of the height/width attributes for dragresize_ie and replaced them with style modifications to line up with the behavior on other browsers.

## Test
This fix has been tested on Internet Explorer 11, the main environment were we use dragresize_ie.

Resending this as a fixed PR from https://github.com/liferay/alloy-editor/pull/920

### DIFF
--- a/src/plugins/dragresize_ie.js
+++ b/src/plugins/dragresize_ie.js
@@ -878,10 +878,9 @@
                 // Don't update attributes if less than 10.
                 // This is to prevent images to visually disappear.
                 if (newWidth >= 15 && (newHeight >= 15 || newHeight === 'auto')) {
-                    image.setAttributes({
-                        width: newWidth,
-                        height: newHeight
-                    });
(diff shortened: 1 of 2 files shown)


julien:

julien:
@gregory-bretall Thanks!

PR #927, opened 2018-11-27
Fixes #906 | Implement feature sets

### LINKED ISSUE #906: Feature Request: Feature Sets
AlloyEditor allows the user to extensively configure its behaviour on several levels:
- Selections
- Toolbars
- Buttons

The default configuration, specially the button selection in Toolbars accounts for a "basic editing" experience. 

From https://issues.liferay.com/browse/LPS-83732:

> Why not re-approach features sets as just that. "Sets". Sure offer the default tools you need to get by (including text alignment) but then offer the user the option to expand their tool "set" so if there's a feature that they need it's a click away for them.
>
> * Basic editing
> * Technical writing
> * Conversational writing
> * "All" ... when the user selects all they get every tool available in Alloy Editor

julien:


jbalsas:
Hey @julien, could you post a bit of an explanation of the proposal here? What feature sets are we enabling by default? What do they contain? Couple of screenshots? 😉 

/cc @duracell80

julien:
Hey @jbalsas sure.

The goal of this change is to be able to use more buttons by using the new `mode` property.

For the moment this is still a proposal, but if you were to configure the editor to use the `full` mode like this:

```javascript
  mode: 'full'
});
``` 

You'd see something like this:

ae_full_toolbar

With the `full` mode these are the buttons that are available:

- styles
- bold 
- italic
- underline
- strike
- link
- Font
- FontSize
- TextColor
- BGColor
- paragraphAlign
- ul
- ol
- h1
- h2
- indentBlock
- outdentBlock
- code
- quote
- removeFormat

We've also added a `separator` element which can be used to display a simple `|` separator between a set of buttons.

You can group buttons, so that each group is placed in a "row" inside the toolbar, this is done by listing the buttons you want in an array, as you can see [here](https://github.com/liferay/alloy-editor/blob/800fd6d637b3314625a4a0d425a3495a2a9161a8/src/ui/react/src/selections/selections.js#L20-L30) 

If no `mode` option is passed, you'd still see the toolbar with the default set of buttons

screen shot 2018-11-27 at 17 03 36

That's all for now, we can think about useful new modes to make users even happier.

jbalsas:
Hey @julien, just curious... how did you updated the ckeditor plugins?

src/ui/react/src/adapter/main.js
```diff
@@ -186,6 +186,16 @@
...
 
+        /**
+         * And object, containing all currently registered elements in AlloyEditor.
+         *
+         * @memberof AlloyEditor
+         * @property Elements
+         * @type {Object}
+         * @static
+         */
+        Elements: {},
```
jbalsas: What's this for?

src/ui/react/_src.js
```diff
@@ -20,6 +20,7 @@ module.exports = {
         'ui/react/src/adapter/alloy-editor.js',
         'ui/react/src/components/base/*.js*',
         'ui/react/src/components/buttons/*.js*',
+        'ui/react/src/components/elements/*.js*',
```
jbalsas: Why create a new elements section?

src/ui/react/src/components/base/toolbar-buttons.js
```diff
@@ -31,12 +58,12 @@
 
             var toolbarButtons = this.filterExclusive(
                     buttons.filter(function(button) {
-                        return button && (AlloyEditor.Buttons[button] || AlloyEditor.Buttons[button.name]);
+                        return button && (AlloyEditor.Buttons[button] || AlloyEditor.Buttons[button.name] || AlloyEditor.Elements[button]);
```
jbalsas: This is a bit weird. I'd just keep the `separator` inside `buttons` for now, even if it's not technically one. All the components that do build buttons are actually in there even though they're not actually buttons.

I'll simplify this for now and keep the hierarchy to a minimum. Once me work on moving to higher React versions we can put things anywhere and rely on dynamic imports or a smarter way to deal with this.

julien: Ok, I thought having an `Elements` section for the `separator` was more logic because it isn't a button, but now that this is clarified, I'll move it to the `buttons` directory. Thanks for the info!

duracell80:
This is a great start. Code, H1 and H2 can probably be removed as they are in the text styles drop-down, quote could be moved into that drop down. 

Is the Paragraph align a drop down? If so it should really have a down facing caret icon.

On the idea of feature sets this is really needing to be an icon that users can click to cycle feature set modes. As developers we shouldn't set these modes and confine users, they should be able to start with the bar as is right now and then themselves click an icon to cycle to the mode they need at that point in time. An "Expand" chevron to make more icons appear, clicking it again shrinks the bar back to basic tools.

duracell80:
This would be an LPS for Liferay but ... The way the alloy editor is right now it is much more suitable for inline basic edits ... Therefore it shouldn't be presented that way in edit web content or blogs or threads for example or anywhere else.

The bar as is now should be for live edits on the page without having to go "edit web content". Let's say when the user "edits web content" then the UI should not be this context bar it should be maore of a traditional document editor. The inline editor mode shouldn't be used for full feature sets. That's the root problem with trying to fit all those icons into inline mode ... It can't be done.

julien:
Hey @jbalsas,

I resent a pull request here https://github.com/liferay/alloy-editor/pull/930

/cc @duracell80

PR #930, opened 2018-11-28
Fixes #906 | Implement feature sets

### LINKED ISSUE #906: Feature Request: Feature Sets

julien:


julien:
Adding this as a note for later:

- Apparently Saucelabs results are random: 
  sometimes restarting the build (with no changes) marks the build as successful, 
  sometimes it marks the build as failed.

- We should probably remove 'colors' from the karma configuration for saucelabs as it makes it impossible to read ... the logs are full of escape codes (`[34m[` and similiar)

jbalsas:
Hey @julien, can we split this into at least 2 commits: 
- Update CKEditor
- Implement Feature Sets

Thanks! ;)

PR #957, opened 2019-01-17
Add Prettier

wincent:
We should merge something like this when things have settled down a bit and some of the currently in-flight PRs have landed, but this is a preview of how things could look.

As noted in the commit message, this is manual for now (`npm run pretty`): if we like it, we can enforce formatting with then linter and in the CI.

Related: https://issues.liferay.com/browse/IFI-411


.prettierrc.json
```diff
@@ -0,0 +1,7 @@
+{
+  "bracketSpacing": false,
+  "jsxBracketSameLine": true,
+  "singleQuote": true,
+  "tabWidth": 4,
+  "useTabs": true
+}
```
wincent: I tried to pick settings that seem to match the (kind of) existing desired style. Do we have a definitive style guide for this anywhere, or a repo where we're already using Prettier with some sort of org-approved settings?

src/lang/af.js
```diff
@@ -1 +1,77 @@
...
\ No newline at end of file
+AlloyEditor.Strings = {
```
wincent: All these lang files look auto-generated, so perhaps they should be excluded (ie. with a `@noformat` annotation in the header).

julien: Yes you're totally right, these are generated.

.prettierrc.json
julien: Yes, we're using in [clay](https://github.com/liferay/clay.git). 

I guess [here](https://github.com/liferay/clay/blob/master/package.json#L15) and maybe [here](https://github.com/liferay/clay/blob/master/.eslintrc) could be a good place to look.

wincent:
OK, so I refreshed this on top of the current `HEAD` of the "2.x-develop" branch. I changed the package.json script names to match the corresponding names in Clay, and I added a `.prettierignore` to ignore the generated files in "src/lang/".

PR #960, opened 2019-01-21
Remove unused gulp-empty dependency

wincent:
Originally submitted as part of #958. Splitting it off to see if Travis accepts a smaller change.

PR #961, opened 2019-01-21
Remove unused gulp-empty dependency

wincent:
Pushing this because for Travis didn't run against #960.


wincent:
Aaaand Travis just decided to run against that other PR after all.

PR #958, opened 2019-01-18
Deal with security issues identified by npm audit

julien:

PR #962, opened 2019-01-21
Remove never-used gulp-empty dependency

wincent:
Added in aa2f8f76b25262982253e3e8a394c7ffdb1284ae and looks like it was
never used, not even in that commit.

Splitting this off from #958 to see if a smaller commit passes Travis.


wincent:
Ok, that passed. My number one suspect is now the "karma-sauce-launcher" package. I'm going to re-roll the other PR without it and then tackle it separately.

PR #970, opened 2019-01-22
Fixes #969 - Don't commit the "dist" contents to the Git repo

### LINKED ISSUE #969: Clean up dist folder
We've been (sporadically) committing the built files in the dist folder for a long time now, which is a non-standard way of operating and probably isn't working so well (given that we don't always remember to update dist). I want to take a look at removing it from the repo and adding some safeguards to make sure that what we produce with `npm publish` is correct.

wincent:
Just putting this here for discussion of the merits of committing/not-committing this stuff to the repo. If we decide to do this, I would want to potentially make some other changes on top of this.


wincent:
(Not for actually merging, just for discussion; cc @julien)

PR #974, opened 2019-01-23
Update karma-sauce-launcher as recommended by `npm audit`

wincent:
Before:

    found 10 vulnerabilities (4 low, 5 high, 1 critical) in 17824 scanned packages

After:

    found 8 vulnerabilities (3 low, 4 high, 1 critical) in 17615 scanned packages

`npm run dev && npm run test` still green after this change is applied.

Note that I had to change one of the Sauce Labs config settings in order to get this to pass in Travis; without this change, CI runs were blowing up with:

```
ERROR [SaucelabsLauncher]: Error: Error: no ssl bump domains(s) not
valid, invalid domain(s): -B all
```

### DIFF
--- a/karma-saucelabs.js
+++ b/karma-saucelabs.js
@@ -68,6 +68,7 @@ const defaultConfig = Object.assign(Common, {
 		recordScreenshots: false,
 		startConnect: true,
 		connectOptions: {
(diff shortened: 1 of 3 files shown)


wincent:
Woot! It worked. 🎉

PR #977, opened 2019-01-23
Remove unused font related resources and tasks

julien:


scripts/build/gulp/tasks/css.js
```diff
@@ -19,7 +19,6 @@ function errorHandler(error) {
 
 gulp.task('build-css', function(callback) {
 	runSequence(
-		'generate-fonts',
 		'sass2css',
 		'join-css',
 		'clean-fonts',
```
wincent: Can we get rid of clean-fonts as well?

```suggestion
```

julien: Yes, good catch

julien:
@wincent yes, we might as well get rid of that too. Thanks!

PR #978, opened 2019-01-23
Remove unnecessary demo app templating

wincent:
The only difference between the prod build and dev build is that the former uses the minified JS and CSS bundles. In practice, the "main" entry point in our package.json points at the non-minified build, so it's appropriate that we use the same build in the demo.

Motivation: this simplifies the build and removes a dependency, which will make it easier to move to Gulp v4.

Test plan: `npm run dev` and `npm run start` and confirm demo works.


julien:

PR #987, opened 2019-01-23
Use a more recent Node in Travis

wincent:
We were pegged at 8, but let's see if we have any more stability by
using the latest stable release at all times:

https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#specifying-nodejs-versions


wincent:
Let's see if this improves our luck?

PR #955, opened 2019-01-16
Migrate away from reactCreateClass

wincent:
For: https://issues.liferay.com/browse/IFI-409

This PR ports us away from the four places where we're using `createReactClass`.

PR #996, opened 2019-01-24
Fix for various buttons

julien:


src/components/uibridge/button.jsx
```diff
@@ -40,9 +40,7 @@ import React from 'react';
 			ButtonBridge = class extends React.Component {
 				static displayName = buttonName;
 
-				statics = {
-					key: buttonName,
-				};
+				static key = buttonName;
```
wincent: Ah sorry, this was me in #955. I had checked the documentation for `create-react-class` but it doesn't really exist. I thought `statics` was just an inappropriately named *instance* variable, but it actually creates static properties on the class (of course!) 🤦‍♂️

PR #1005, opened 2019-01-29
Fix lint in src/components/base/widget-exclusive.js

wincent:
```
  39:29  error  'nextProps' is defined but never used. Allowed unused args must match /^_/  no-unused-vars
```

We were calling `super()` without propagating the args. Note that the way we are using subclassing in our HOCs is not a very orthodox approach; "The React Way" is to favor composition over inheritance.

Usual test plan: `npm run dev && npm run test && npm run start`; check demo.


julien:

PR #1018, opened 2019-01-29
Fix lint in uibridge/menu.jsx (guard-for-in)

wincent:
Wrote some old-school code to fix this:

```
src/components/uibridge/menu.jsx
  79:5  error  The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype  guard-for-in
```

FWIW, the "clever" way to do this with `Object.entries().forEach()` ends up being less readable because of the need to use a non-arrow function and pass in `this`.

Usual tests: `npm run dev && npm run test && npm run start` and look at demo.

### DIFF
--- a/src/components/uibridge/menu.jsx
+++ b/src/components/uibridge/menu.jsx
@@ -77,7 +77,9 @@
 			 */
 			editor.addMenuItems = function(definitions) {
 				for (let itemName in definitions) {
-					this.addMenuItem(itemName, definitions[itemName]);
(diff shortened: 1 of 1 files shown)


wincent:
> I also like the `Object.entries.forEach` but yes it is probably less readable.

Yeah, in this case it ends up looking like:

```javascript
Object.entries(definitions).forEach(function ([itemName, definition]) {
  this.addMenuItem(itemName, definition);
}, this);
```

which is a bit 🤢

😂

PR #1019, opened 2019-01-29
Fix 9 remaining "guard-for-in" lints

wincent:
```
src/plugins/dragresize_ie.js
  330:6   error  The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype  guard-for-in
  554:4   error  The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype  guard-for-in

src/plugins/dragresize_ie11.js
    143:4   error  The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype  guard-for-in
    384:6   error  The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype  guard-for-in
  1038:4   error  The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype  guard-for-in

src/plugins/dragresize.js
  238:4   error  The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype  guard-for-in
  327:4   error  The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype  guard-for-in
  334:4   error  The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype  guard-for-in
  340:4   error  The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype  guard-for-in
```

One other one is fixed in: https://github.com/liferay/alloy-editor/pull/1018

This is pretty damn paranoid - some of these are operating on object literals, so the only way the guard could prevent garbage from getting through would be if somebody had defined enumerable properties on `Object.prototype` - but it is better to be safe than sorry I guess. And seeing as at least a couple of these are in modules built for IE, I'm being super conservative and just applying the change suggested by the linter.

### DIFF
--- a/src/plugins/dragresize.js
+++ b/src/plugins/dragresize.js
@@ -236,7 +236,9 @@
 			});
 
 			for (let n in handles) {
-				container.appendChild(handles[n]);
+				if (handles.hasOwnProperty(n)) {
+					container.appendChild(handles[n]);
+				}
 			}
 		},
 		createHandle: function(name) {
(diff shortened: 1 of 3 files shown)


julien:

PR #1020, opened 2019-01-29
Fix 3 lints in src/core/uicore.js

wincent:
```
src/core/uicore.js
   99:54  error  'event' is defined but never used. Allowed unused args must match /^_/  no-unused-vars
  135:5   error  Unexpected var, use let or const instead                                no-var
  154:34  error  'event' is defined but never used. Allowed unused args must match /^_/  no-unused-vars
```

Usual tests: `npm run dev && npm run test && npm run start` and check in demo that things work and no errors in console.

### DIFF
--- a/src/core/uicore.js
+++ b/src/core/uicore.js
@@ -96,7 +96,7 @@
 				}
 			}, uiTasksTimeout);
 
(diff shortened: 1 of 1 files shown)


julien:

PR #1026, opened 2019-01-29
Fix bad identifier in button-embed-video-edit.jsx (no-undef)

wincent:
```
src/components/buttons/button-embed-video-edit.jsx
   99:5   error  'videoUrl' is not defined                              no-undef
```

Also looks like a legit bug.

Test plan: `npm run dev && npm run test && npm run start` and hit demo.

Related: https://github.com/liferay/alloy-editor/issues/990

PR #1025, opened 2019-01-29
Fix bad variable name in button-link-browse.jsx (no-undef)

wincent:
```
src/components/buttons/button-link-browse.jsx
  75:31  error  'LinkBrowse' is not defined  no-undef
```

This one looks to be another legit bug. Yay for the linter!

Test plan: `npm run dev && npm run test && npm run start`, see demo.

Related: https://github.com/liferay/alloy-editor/issues/990

PR #1006, opened 2019-01-29
Remove timeout in tests

julien:

PR #1027, opened 2019-01-29
Fix no-undef lint in src/plugins/dragresize_ie11.js

wincent:
```
src/plugins/dragresize_ie.js
  1410:5   error  'moveRatio' is not defined           no-undef
```

This is another "no-undef" "bug" but not like the ones in #1025 and #1026. Here we have something that is missing a `let` keyword, but it turns out variable is unused anyway, so I just removed it.

Last use was removed in d6145d2c4ad866fbe, I believe.

In any case, no real harm here at runtime, because I believe it just ended up assigning to `window.moveRatio`.

Test plan: `npm run dev && npm run test && npm run start`

### DIFF
--- a/src/plugins/dragresize_ie11.js
+++ b/src/plugins/dragresize_ie11.js
@@ -1415,9 +1415,6 @@
 				moveDiffX = nativeEvt.screenX - startX;
 				moveDiffY = startY - nativeEvt.screenY;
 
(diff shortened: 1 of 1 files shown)


wincent:
Rebased on top of #1006, which should make it easier for this thing to pass Travis.

julien:

Issue #1030, opened 2019-01-29
React warnings when editing photo "alt" tags


wincent:
Repro:

0. Open the demo.
1. Add a photo.
2. Click the "alt" action.

See something like:

```
react-dom.development.js:506 Warning: `value` prop on `input` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components.
    in input (created by _class)
    in div (created by _class)
    in div (created by _class)
    in _class (created by _class)
    in div (created by _class)
    in div (created by _class)
    in _class (created by _class2)
    in div (created by _class2)
    in _class2
warningWithoutStack @ react-dom.development.js:506
warning @ react-dom.development.js:2618
validateProperties$1 @ react-dom.development.js:7206
validatePropertiesInDevelopment @ react-dom.development.js:7414
setInitialProperties @ react-dom.development.js:7664
finalizeInitialChildren @ react-dom.development.js:8725
completeWork @ react-dom.development.js:15817
completeUnitOfWork @ react-dom.development.js:17980
performUnitOfWork @ react-dom.development.js:18178
workLoop @ react-dom.development.js:18190
renderRoot @ react-dom.development.js:18276
performWorkOnRoot @ react-dom.development.js:19165
performWork @ react-dom.development.js:19077
performSyncWork @ react-dom.development.js:19051
interactiveUpdates$1 @ react-dom.development.js:19320
interactiveUpdates @ react-dom.development.js:2169
dispatchInteractiveEvent @ react-dom.development.js:4880
react-dom.development.js:506 Warning: A component is changing an uncontrolled input of type text to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components
    in input (created by _class)
    in div (created by _class)
    in div (created by _class)
    in _class (created by _class)
    in div (created by _class)
    in div (created by _class)
    in _class (created by _class2)
    in div (created by _class2)
    in _class2
```

PR #1033, opened 2019-01-30
Eliminate React warnings when editing image alt attribute

wincent:
We had this console spew:

```
react-dom.development.js:506 Warning: `value` prop on `input` should
not be null. Consider using an empty string to clear the component or
`undefined` for uncontrolled components.
```

With this change, we no longer pass a null-ish value as and the console is clean.

Test plan: `npm run dev && npm run test && npm run start`; then in demo, click on image, click on alt button, see console is clean and setting an alt still works.

Closes: https://github.com/liferay/alloy-editor/issues/1030
Internal: https://issues.liferay.com/browse/IFI-435


julien:

PR #1039, opened 2019-01-30
Fix 11 lints in src/plugins/tabletools.js

wincent:
```
src/plugins/tabletools.js
   53:5  error  Unexpected var, use let or const instead   no-var
  167:4  error  Unexpected var, use let or const instead   no-var
  196:4  error  Unexpected var, use let or const instead   no-var
  212:9  error  Unexpected var, use let or const instead   no-var
  320:8  error  Unexpected var, use let or const instead   no-var
  329:4  error  Unexpected var, use let or const instead   no-var
  370:8  error  Unexpected var, use let or const instead   no-var
  371:9  error  Unexpected var, use let or const instead   no-var
  591:8  error  Unexpected var, use let or const instead   no-var
  709:4  error  Unexpected var, use let or const instead   no-var
  803:8  error  'lang' is assigned a value but never used  no-unused-vars
```

`lang` is safe to remove because the last references to it look like they were removed in c2d3248c3cf6f883c01b548f0caad2860862ae66 ("Remove ui-related code from plugins", 2015).

The others I changed to `let` or `const` as appropriate. This code leans too heavily on `let` semantics in general, because values are reassigned and accessed sometimes quite large distances away.

One particularly nasty case was a `for (var i = 0, rows = ...)`. This `rows` var is initialized in the loop and then accessed quite a bit further down outside the loop. For that one I had to hoist the `let rows` higher up in the scope. This is a bit of a foot-gun, and I only caught it by testing.

Test plan: `npm run dev && npm run test && npm run start` and test demo
(specifically tables).

### DIFF
--- a/src/plugins/tabletools.js
+++ b/src/plugins/tabletools.js
@@ -50,7 +50,7 @@
 				if (nearestCell) retval.push(nearestCell);
 			} else {
 				let walker = new CKEDITOR.dom.walker(range);
-				var node;
+				let node;
 				walker.guard = moveOutOfCellGuard;
 
 				while ((node = walker.next())) {
@@ -164,7 +164,7 @@
 
 		let newRow = doc.createElement('tr');
 		for (let i = 0; cloneRow[i] && i < width; i++) {
-			var cell;
+			let cell;
 			// Check whether there's a spanning row here, do not break it.
 			if (
 				cloneRow[i].rowSpan > 1 &&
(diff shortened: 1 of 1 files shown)


src/plugins/tabletools.js
```diff
@@ -366,17 +366,19 @@
 
 		let rowsToDelete = [];
 
+		let rows;
+
 		// Figure out selected cells' column indices.
-		for (var i = 0, rows = map.length; i < rows; i++) {
-			for (var j = 0, cols = map[i].length; j < cols; j++) {
+		for (let i = 0, rows = map.length; i < rows; i++) {
+			for (let j = 0, cols = map[i].length; j < cols; j++) {
```
wincent: Note this `cols` doesn't suffer the same problem as `rows` in the line above. `cols` is only accessed in the loop condition and not outside of it.

julien:

PR #1042, opened 2019-01-30
Fix 6 lints in src/plugins/tableresize.js (no-var)

wincent:
```
src/plugins/tableresize.js
   98:4  error  Unexpected var, use let or const instead  no-var
   99:4  error  Unexpected var, use let or const instead  no-var
  100:4  error  Unexpected var, use let or const instead  no-var
  275:5  error  Unexpected var, use let or const instead  no-var
  377:3  error  Unexpected var, use let or const instead  no-var
  412:3  error  Unexpected var, use let or const instead  no-var
```

Test plan: `npm run dev && npm run test && npm run start` and test table
functionality in demo.

### DIFF
--- a/src/plugins/tableresize.js
+++ b/src/plugins/tableresize.js
@@ -95,9 +95,9 @@
 			pillarIndex += td.$.colSpan || 1;
 
 			// Calculate the pillar boundary positions.
-			var pillarLeft;
-			var pillarRight;
-			var pillarWidth;
+			let pillarLeft;
+			let pillarRight;
+			let pillarWidth;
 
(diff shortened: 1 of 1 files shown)


src/plugins/tableresize.js
```diff
@@ -163,6 +163,8 @@
 
 	function ColumnResizer(editor, pillar) {
 		let document;
+		let isResizing;
+		let move;
```
wincent: These are another nasty example. Assigned somewhere in the middle of the function, hoisted to function scope, and used in some distant places.

julien:

Issue #1037, opened 2019-01-30
Investigate use of React hook to replace HOCs


wincent:
Our HOCs are pretty non-conventional because they involve wrapping our components using subclasses. In React, it is more natural to favor composition over inheritance.

In many cases, we wrap multiple times, and then end up doing (deprecated `ReactDOM.findDOMNode`) to reach down and find the underlying DOM node to do something like `focus()` it.

So, to get rid of those `findDOMNode` calls, we need to use refs... and to propagate the refs in the context of HOCs, we need to forward them... but before we do that, we should look at whether we could replace some of these HOCs with hooks, which have just exited experimental status. (This is what hooks were pretty much designed for: encapsulating common functionality that can be "mixed-in" and combined in various ways.)

Internal ticket: https://issues.liferay.com/browse/IFI-437

wincent:
Ok, we're calling this investigation completed. As summarized [here](https://github.com/liferay/alloy-editor/pull/1092#issuecomment-464702431):

> Ok, we discussed the merits of this approach and reached these basic conclusions:
> 
> * Hooks look promising for this use case.
> * There is likely a lot of work remaining to complete the migration, and some of it is difficult.
> * With 2.0 on the verge of release, we don't want to delay or complicate things by engaging in a massive rearchitecting, so we'll pause this work for now; we might want to revisit it for 2.1 though.
> 
> So I pushed the WIP to [the "research/hooks" branch](https://github.com/liferay/alloy-editor/tree/research/hooks) and we can take things from there or at least take some inspiration from it, next time we visit this problem space.
> 
> Closing this one for now.

PR #1045, opened 2019-01-30
Exploratory refactor of HOCs (do not merge)

wincent:
As explained in #1037, there are some issues with the way we are using HOCs:

- We have multiple layers of wrapping and drill down through them using `ReactDOM.findDOMNode()`, which is deprecated.
- Forwarding refs through in order to avoid `findDOMNode()` is difficult because of the layers of indirection and the unorthodox wrapping model (which uses wrapping instead of composition).

As an exploratory refactoring, I am going to see if I can simplify some of this.

Starting with the first commit: "Remove super() calls from ButtonKeystroke HOC"

I am looking at teasing apart some of these HOCs. Starting with this one because it is only used in a few places. This one calls "super" but looking at the 6 callsites, none of them implement the overridden methods. So I am going to remove the `super()` calls to make it clear that there is no interaction with the superclass and these can probably be replaced with more conventional classes that don't require inheritance.

For reference, the 6 callsites are:

- src/components/buttons/button-embed.jsx
- src/components/buttons/button-link-browse.jsx
- src/components/buttons/button-link.jsx
- src/components/buttons/button-italic.jsx
- src/components/buttons/button-bold.jsx
- src/components/buttons/button-underline.jsx

And `ButtonKeystroke` is not exported anywhere as far as I can tell, so we can be confident that there are no unknown callers.

Related: https://github.com/liferay/alloy-editor/issues/1037

### DIFF
--- a/src/components/base/button-keystroke.js
+++ b/src/components/base/button-keystroke.js
@@ -16,10 +16,6 @@ export default WrappedComponent =>
 		 * @method componentWillMount
 		 */
 		componentWillMount() {
-			if (Lang.isFunction(super.componentWillMount)) {
-				super.componentWillMount();
-			}
-
 			let nativeEditor = this.props.editor.get('nativeEditor');
 			let keystroke = this.props.keystroke;
 
@@ -58,10 +54,6 @@ export default WrappedComponent =>
 		 * @method componentWillUnmount
 		 */
 		componentWillUnmount() {
-			if (Lang.isFunction(super.componentWillUnmount)) {
-				super.componentWillUnmount();
-			}
-
 			this.props.editor
 				.get('nativeEditor')
 				.setKeystroke(


wincent:
So I was wrong about call sites not implementing the overridden methods. `ButtonKeystroke` sometimes wrap `ButtonStyle`, which *does* implement the method. So I'm actually going to start there instead. There are 11 components wrapped in `ButtonStyle`, I think. I'm going to check whether they implement `component{Did,Will}Unmount` an continue that way if I can.

julien:

wincent:
This one is going to need a bit of work before it actually functions; and that led me to see that switching to React context would be a good preliminary step.

jbalsas:
Switch to Context merged! -> https://github.com/liferay/alloy-editor/pull/1055

wincent:
Closing this one in favor of #1092.

PR #1055, opened 2019-01-31
Start using React context to share editor instance

wincent:
Instead of passing the editor through props everywhere, we can get it out of props and into context. This will make dealing with #1037 easier, I hope, because it will be one less prop we have to worry about propagating through our HOCs; but I think it makes sense to do this anyway because the editor is a "global" value ("global" per editor instance, that is), which is exactly the use case that context exists for.

First commit shows using context in a class component at the top of the hierarchy. Second commit shows using context in a functional component in a leaf at the bottom of the hierarchy. I'll then proceed to add more commits until the whole tree is ported. At each step, my test plan is `npm run dev && npm run test && npm run start` and checking the demo.


src/components/buttons/button-accessibility-image-alt.jsx
```diff
@@ -27,10 +30,10 @@ class ButtonAccessibilityImageAlt extends React.Component {
 		editor: PropTypes.object.isRequired,
 	};
 
-	constructor(prop, context) {
-		super(prop, context);
+	constructor(prop) {
+		super(prop);
```
wincent: Constructors don't receive `context` (that was part of the legacy context API), and in any case, I don't think we were ever using it here.

wincent:
I think this is ready for review now @julien.

wincent:
Rebased onto current `2.x-develop`.

wincent:
Tested this pretty thoroughly in demo and seems to work in portal too:

new_web_content_-_liferay

src/adapter/editor-context.js
```diff
@@ -0,0 +1,27 @@
...
+	const componentWithContext = props => (
+		<EditorContext.Consumer>
+			{context => <Component {...props} context={context} />}
+		</EditorContext.Consumer>
+	);
+	componentWithContext.key = Component.key;
+	return componentWithContext;
+};
+
+export default EditorContext;
```
jbalsas: Not sure where we stand on this from our previous conversations on other projects... not to worry here, but should we go with named exports everywhere? Thought that was the consensus 🤷‍♀️

src/components/base/toolbar-buttons.js
```diff
@@ -1,16 +1,19 @@
...
  *
  * @class ToolbarButtons
  */
 export default WrappedComponent =>
-	class extends WrappedComponent {
+	class ToolbarButtons extends WrappedComponent {
+		static contextType = EditorContext;
+
 		/**
-		 * Analayses the current selection and returns the buttons or button groups to be rendered.
```
jbalsas: You can't unseen this 😂

src/components/buttons/button-accessibility-image-alt.jsx
```diff
@@ -14,23 +14,16 @@ import ReactDOM from 'react-dom';
...
-		/**
-		 * The editor instance where the component is being used.
-		 *
-		 * @instance
-		 * @memberof ButtonAccessibilityImageAlt
-		 * @property {Object} editor
-		 */
-		editor: PropTypes.object.isRequired,
-	};
+	static propTypes = {};
```
jbalsas: Just out of curiosity, what's the benefit of an empty `proptypes` definition? Looks like we're not validating `propTypes` in many places... should we file an issue to track down what we want to do about it?

wincent: I don't think there's any real "benefit". I just wanted to make the minimal change, which was removing the property. You probably won't be surprised to know that I would rather see us using static types and just get rid of runtime prop-type checking.

src/adapter/editor-context.js
wincent: Previous discussion was on https://github.com/liferay/clay/pull/1453 — summarizing:

- That PR just uses a default export.
- Some articles were shared arguing in favor of named exports.
- @julien expressed a preference for named exports.
- You noted that, at least in Clay, we pretty strictly follow a one-export-per-module, making default exports simple.
- PR got merged, possibly without enough explicit agreement on what exactly we were agreeing to.

I'm open to sticking to default exports or named exports, and I think the one-export-per-module guideline is a good rule of thumb. Being consistent about it and picking "an" option is probably more valuable than picking the "right" option. It's not too late to change our minds if we think that one was a mistake.

src/components/buttons/button-accessibility-image-alt.jsx
wincent: Followed up in #1084.

PR #1064, opened 2019-02-05
Modify github issue and pull request templates:

julien:


julien:
@wincent this is not urgent at all but I thought it might be better if you had a look at this when you can.
Thanks!

Issue #1068, opened 2019-02-06
Still requesting font assets after migration to SVG-based icons


wincent:
I was of the understanding that we shouldn't be requesting fonts any more since migrating to SVG-based icons. But I see the following in the console on the current HEAD of the 2.x-develop branch:

```
GET http://localhost:9000/alloy-editor/assets/fonts/alloyeditor-ocean.woff net::ERR_ABORTED 404 (Not Found)
GET http://localhost:9000/alloy-editor/assets/fonts/alloyeditor-ocean.ttf net::ERR_ABORTED 404 (Not Found)
```

Not sure if there is some other reason (other than icons) why we might be requesting a font that isn't there, but creating this issue in order to look in to why this error is presenting itself.

PR #1069, opened 2019-02-06
Don't ship font-related CSS any more

wincent:
We moved to SVG icons from font-based ones in d1b061729f9ca4f9476059c, but we have continued to include the font-related CSS in our build, which leads to console spam like this when you do certain things:

```
GET http://localhost:9000/alloy-editor/assets/fonts/alloyeditor-ocean.woff net::ERR_ABORTED 404 (Not Found)
GET http://localhost:9000/alloy-editor/assets/fonts/alloyeditor-ocean.ttf net::ERR_ABORTED 404 (Not Found)
```

(For example, if you hit "+" in the LHS margin and then hit the "+" icon in the pop-up toolbar.)

We should be able to omit this safely. There are still some straggling references to `ae-icon-*` classes in the codebase (apart from the `ae-icon-svg-*` ones, which are legit), but removing these will be the subject of another pull.

Tested plan: `npm run build && npm run test && npm run start`

Fixes: #1068

### DIFF
--- a/scripts/build/gulp/tasks/css.js
+++ b/scripts/build/gulp/tasks/css.js
@@ -38,23 +38,12 @@ function getJoinTasks() {
 
 	return skins.map(function(skin) {
 		const skinFileName = 'alloy-editor-' + skin + '.css';
-		const skinFontFileName = 'alloyeditor-font-' + skin + '.css';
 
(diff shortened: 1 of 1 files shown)


julien:

Issue #1070, opened 2019-02-06
Create AlloyEditor React component


julien:
Since we're making a new demo site with Gatsby, it would be great to have an `AlloyEditor` component that we can render directly in JSX.

wincent:
Just to reflect here some discussion we had offline: one of the motivations here is to solve the pain we've had on the new Alloy Editor website of using the editor there in example/playground format. @julien was showing me how it is pretty hacked together at the moment (defining globals on `window` and having to move folders and directories around because the code — or the build, I am not sure which — is making some location-specific assumptions). The goal would be to make consumption of the editor trivially easy.

PR #1082, opened 2019-02-08
Fixes #1081 - Make dev environment hot-reloading work

### LINKED ISSUE #1081: Fix hot-reloading in development environment
The webpack dev server is recompiling when a change is made to the source files, but the browser demo doesn't pick it up without a full page refresh.

wincent:
We need the "publicPath" setting to make this work. With this change, `npm run start` prints this:

```
ℹ ｢wds｣: Project is running at http://localhost:9000/
ℹ ｢wds｣: webpack output is served from /alloy-editor
ℹ ｢wds｣: Content not from webpack is served from ./dist
```

Note the "served from /alloy-editor" bit. (Prior to this commit is instead said "served from /".) This allows the "alloy-editor/alloy-editor-all.js" script that our demo index.html references to be hot-reloaded.

Test plan: `npm run dev && npm run start`; hit demo and see it working, then add a `console.log`, see webpack recompile, and see the browser reload the module without any intervention (output appears in console).

Closes: https://github.com/liferay/alloy-editor/issues/1081

### DIFF
--- a/webpack.dev.js
+++ b/webpack.dev.js
@@ -14,6 +14,7 @@ const All = merge(common.config, config, {
 	devServer: {
 		contentBase: './dist',
 		port: 9000,
+		publicPath: '/alloy-editor',
 	},
 });
 


jbalsas:
Sweet, this got me thinking wether we could try to get Hot Module Replacement working... maybe https://github.com/liferay/alloy-editor/issues/1070 is a requirement before we can get there?

Merging this in!

wincent:
TBH, not sure how much of a difference there is between hot reloading (what's happening as a result of this PR) and hot module replacement (documented [here](https://webpack.js.org/guides/hot-module-replacement/)). I suspect that there is enough non-React state on the CKEditor side that even with #1070 in place, I have no idea if it will be possible to usefully preserve module state across reloads.

PR #1083, opened 2019-02-08
Update to React v16.8.1

wincent:
Test plan: `npm run build && npm run test && npm run start` and check demo.


jbalsas:
The one with hooks? 😱

PR #1092, opened 2019-02-11
HOC -> hook POC (do not merge)

wincent:
This is a work-in-progress exploration of using hooks in Alloy Editor.

I tried to pick the simplest button (`<ButtonHline />`) and replace the HOCs that it uses with hooks:

- `ButtonCommand` -> `useExecCommand`: This HOC adds an `execCommand` function to the wrapped instance, which translates fairly straightforwardly to a hook that plucks the native editor out of the context and calls `execCommand()` on it. The gotcha here is that it then does an `editor.fire()`, passing in `this` as a parameter. There is probably nothing inside Alloy Editor that relies on this, but we have at least two places in Liferay Portal where we are receiving a component instance via this mechanism and then violating encapsulation by reaching into the `props` member to grab the corresponding `command`:
  - https://github.com/liferay/liferay-portal/blob/4b5542e86a251faa27c29a8d52cd8a04bde141b7/modules/apps/dynamic-data-mapping/dynamic-data-mapping-form-field-type/src/main/resources/META-INF/resources/editor/editor_field.js#L152
  - https://github.com/liferay/liferay-portal/blob/399e1b734a97033f4da9f14dbd676ef24019314d/modules/apps/dynamic-data-mapping/dynamic-data-mapping-form-field-type/src/main/resources/META-INF/resources/metal/Editor/Editor.es.js#L205
- `ButtonStyle` -> `useButtonStyle`: This HOC adds `isActive()` and `getStyle()` functions to the wrapped instance which are then called in various places, including plug-ins, which makes it a bit hard to determine where/what all the usages are. This one translates into a hook that again plucks the native editor out of the context, and uses `useMemo` to avoid repeatedly re-creating a new `CKEDITOR.style` (the HOC used `componentDidMount()` to cache a copy on the component instance).

With these changes, the HR button still works, but clearly further work is needed; this approach is not going to work on buttons which have places relying on the `isActive()` and `getStyle()` functions (note that in this commit I'm not even doing anything with the `isActive` and `style` values returned by the hook).

I'll keep exploring this, but it is looking like a move towards hooks will be a pretty huge change, because:

1. So much of the system assumes it will be working with class component instances.
2. We violate encapsulation of those instances in a number of places (doing things like accessing props and state from the outside).
3. We have a lot of places where we really on implicit assumptions about the value of `this` (which is not useful in the context of functional components).
4. Hooks only work with functional components.

Related:  https://github.com/liferay/alloy-editor/issues/1037


wincent:
@brunobasto: Mentioning you here as a heads up because of a pattern we found in the portal forms code (see links in the PR summary) where we are relying on a component instance being passed as data in an "onActionPerformed" callback, and then we're reaching into the props and plucking out a command.

I don't know whether the code in question is dealing with Alloy Editor (I just found it with a dumb grep for "actionPerformed"), or whether we'll proceed with the direction explored in this PR, but we probably want to move away from inspecting props from the outside in any case because it breaks encapsulation and makes it hard to do refactoring inside the editor without breaking things; we should explore an alternative mechanism for achieving the same ends.

wincent:
Ok, we discussed the merits of this approach and reached these basic conclusions:

* Hooks look promising for this use case.
* There is likely a lot of work remaining to complete the migration, and some of it is difficult.
* With 2.0 on the verge of release, we don't want to delay or complicate things by engaging in a massive rearchitecting, so we'll pause this work for now; we might want to revisit it for 2.1 though.

So I pushed the WIP to [the "research/hooks" branch](https://github.com/liferay/alloy-editor/tree/research/hooks) and we can take things from there or at least take some inspiration from it, next time we visit this problem space.

Closing this one for now.

Issue #1048, opened 2019-01-30
UX issue, trying to "update an existing image"


duracell80:
Stepping outside of our own experience a user is finding it challenging to see how to "update an image" using Alloy Editor. When an image is added and linked it's not so easy to figure out how to replace that image with a new one and indeed replacing the image removes the link that was on the image.

When adding an image this is easy the user uses the "plus button" to activate a bar that contains an image icon. When trying to update that image file with a new one it's not in the users mind to "add" an image, they want to replace it and don't think the plus button would be of use.

Clicking an existing image brings up a bar that is without an image icon. They must click the plus button to open up a different bar to the one that appears when they click the existing image. Additionally actually selecting a new image file wipes off the original link so the user must replace their link too.

See GIF.

![ux-difficult-1](https://user-images.githubusercontent.com/18121278/52012923-743cff80-24a1-11e9-845f-e0f4c1f23141.gif)

jbalsas:
Hey @duracell80, this is certainly a very valid concern. What would you expect... an additional button in the toolbar, maybe with the image or pencil icon? Thinking about this, I assume it might be just as unintuitive to remove the image?

duracell80:
Honestly ... the ideal solution would be to rethink the UX in DXP and have a static toolbar that didn't move or jump or have three separate parts to it (unless the user was live editing).

On this issue though, an icon in the toolbar that appears when a user clicks the existing image would be a good solution. Keeping it the same icon that's in the plus menu would help users identify. The perfect solution would be a right click context menu, that would then have "image properties", "remove image", "update image", "cut/copy/paste".

I think most users if shown CKEditor would be of the opinion ... why can't we have those features? We had those features in 6.1 why did the editor get harder to use?

PR #1096, opened 2019-02-12
Fixes #1048 | Add button to remove images

julien:
The next part of the fix is to add a button to update
images.

### DIFF
--- a/src/adapter/core.js
+++ b/src/adapter/core.js
@@ -99,19 +99,16 @@ extend(
 
 					let extraCommands = this.get('extraCommands');
 
-					const extraCommandKeys = Object.keys(extraCommands);
-					for (let i = 0; i < extraCommandKeys.length; i++) {
-						const commandName = extraCommandKeys[i];
-
+					Object.entries(extraCommands).forEach(([commandName, command]) => {
 						if (editor.commands[commandName]) {
-							continue;
+							return;
 						}
 
 						editor.addCommand(
 							commandName,
 							extraCommands[commandName]
 						);
-					}
+					});
 
 					editable.addClass('ae-editable');
 				}.bind(this)
--- a/src/adapter/core.js
+++ b/src/adapter/core.js
@@ -99,16 +99,19 @@ extend(
 
 					let extraCommands = this.get('extraCommands');
 
-					Object.entries(extraCommands).forEach(([commandName, command]) => {
+					const extraCommandKeys = Object.keys(extraCommands);
+					for (let i = 0; i < extraCommandKeys.length; i++) {
+						const commandName = extraCommandKeys[i];
+
 						if (editor.commands[commandName]) {
-							return;
+							continue;
 						}
 
 						editor.addCommand(
 							commandName,
 							extraCommands[commandName]
 						);
-					});
+					}
 
 					editable.addClass('ae-editable');
 				}.bind(this)
--- a/src/adapter/core.js
+++ b/src/adapter/core.js
@@ -10,6 +10,8 @@ import Base from '../oop/base';
 import Selections from '../selections/selections';
 import UI from '../components/main.jsx';
 
+import {removeImageCommand} from '../commands';
+
 import React from 'react';
 import ReactDOM from 'react-dom';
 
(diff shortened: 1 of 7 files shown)


src/components/buttons/button-remove-image.jsx
```diff
@@ -0,0 +1,71 @@
...
+
+	static key = 'removeImage';
+
+	/**
+	 * @inheritDoc
+	 */
+	constructor(props) {
+		super(props);
+
+		const nativeEditor = props.editor.get('nativeEditor');
```
wincent: To be consistent with the other buttons, we should be getting `editor` from the context and no the props. But the gotcha is that context is not available in a constructor without using our `EditorContext.toProps` hack, which obliges you to access the editor via `props.context.editor`.

Having said that, you don't see to need the editor here anyway, as it is getting passed in to the `exec` callback below.

julien: Ok, being consistent always sounds good, so I'll use our `EditorContext.toProps` method ... actually I need the `nativeEditor` to be able to call `addCommand` on it later.

I also noticed that this is usually done through plugins but I thought that for this, it wasn't really worth creating a plugin.

julien: @wincent I made some changes, please let me know what you think when you can (no rush)

 I can send another PR to avoid "polluting" the repo's history and logs, but I thought it would be nice to iterate on this until we think it's the right way to go.

src/adapter/core.js
```diff
@@ -95,6 +97,22 @@ extend(
...
+						if (editor.commands[commandName]) {
+							continue;
+						}
+
+						editor.addCommand(
+							commandName,
+							extraCommands[commandName]
+						);
+					}
+
```
wincent: Could also use `Object.entries` here:

```javascript
Object.entries(extraCommands).forEach(([commandName, command]) => {
  if (!editor.commands[commandName]) {
    editor.addCommand(commandName, command);
  }
});
```

src/components/buttons/button-remove-image.jsx
```diff
@@ -0,0 +1,37 @@
...
+		command: 'removeImage',
+	};
+
+	static key = 'removeImage';
+
+	/**
+	 * @inheritDoc
+	 */
+	render() {
+		const cssClass = `ae-button ${this.getStateClasses()}`;
```
wincent: Looking forward to getting away from this pattern (where we call a method that we don't define on the class instance but inherit magically due to our HOC's use of `extends`).

src/lang/en.js
```diff
@@ -81,5 +81,6 @@ AlloyEditor.Strings = {
 	"success": "Success",
 	"table": "Insert Table",
 	"videoPlaybackDisabled": "Video playback is disabled during edit mode",
-	"warning": "Warning"
```
wincent: If I am reading the Gulp build right, I think we don't update this file by hand. Instead, we update `src/assets/lang/en.json`, and then when you run Gulp, it overwrites `src/lang/en.js`.

wincent: (Currently working on updating CONTRIBUTING documenting how to do this, and I think I'll add a warning to not update a generated file as well.)

julien: Yes you're right ... I updated the wrong file 😭

I'll revert this last commit and update the correct file, but yes updating the CONTRIBUTING document seems like a good idea.

wincent: Updated in https://github.com/liferay/alloy-editor/pull/1099

Next thing I'll do is add a "do not edit" header at the top and to be honest I may even move them somewhere where it is very obvious that they shouldn't be edited, like in a directory called `src/__generated__` or something...

julien: Nice, yes that `src/__generated__` directory sounds like a good idea

src/adapter/core.js
julien: I always forget about IE11 

I'll switch back to `Object.keys` because I was looking at the saucelabs logs I see:

```
IE 11.0.0 (Windows 10.0.0) ButtonBold "before all" hook FAILED
	Object doesn't support property or method 'entries' (:121)
```

src/lang/en.js
wincent: Ok, went ahead and did that in https://github.com/liferay/alloy-editor/pull/1107

wincent:
Merging this so it doesn't get stale with respect to the parallel changes going on with the language files (eg. #1107).

PR #1101, opened 2019-02-13
Fixes #1100 - Unbreak language asset builds

### LINKED ISSUE #1100: Fix broken language asset builds
These aren't working since 2b0f4c240, and our experience in #1096 shows us that we should also make it harder to edit the wrong file.

wincent:
Three things:

1. Updates were broken in 2b0f4c2, which added a regex which couldn't possibly match any of the files because none of them are of the form "a.js". The new regex added in this commit matches "en.js", "en-au.js" etc. We could make it stricter, but I don't think that's necessary.
2. Add warning header to generated language files to make it harder to accidentally edit the wrong file, which is something that happened in #1096.
3. Actually update the files.

Closes: #1100


wincent:
This is a fun one, @jbalsas. I wonder what kind of 🤡 reviewed #997. Looks like it was me! 😂

julien:

PR #1106, opened 2019-02-14
Remove explicit ".js" extension in import statements & requires

wincent:
The extension is unnecessary and we were not using it consistently, so let's remove it.

Test plan: `npm run build && npm run test && npm run start` and check demo.


julien:

PR #1107, opened 2019-02-14
Move generated language files to src/__generated__/lang

wincent:
As discussed [here](https://github.com/liferay/alloy-editor/pull/1096#discussion_r256485216), the idea is to make it really unlikely that we'll ever accidentally edit a generated JS file instead of the source JSON file.
    
Test plan: `npm run build && npm run test && npm run start`, check demo. See that `npm run build:assets` updates the generated files when a change is made to the source JSON files.


julien:

PR #1111, opened 2019-02-14
Fix #1109 auto-embedding of pasted video (etc) URLs

### LINKED ISSUE #1109: Fix auto-embedding of pasted video URLs
If you paste in a YouTube URL, it is supposed to turn into an embed, but instead it turns into a link due to the order in which paste events are handled. I have a fix in mind for this so will send a PR shortly. I imagine this has probably been broken for a while.

wincent:
The "embed" plugin checks to see whether pasted text matches a regex for strings starting with "http://" (or "https://"), but the "autolink" plug-in also listens to paste events and ends up transforming any links into anchor tags before the "embed" plug-in sees them. This means that the regex doesn't match and the embed will never work.

In this commit we add a priority mechanism in order to allow the "embed" plug-in to run first.

The gotcha: when we paste a link, the "selectionChange" handler runs in the "embed" plug-in and ends up firing an "editorInteraction" event with the new selection, and that means we see the link toolbar pop-up. Previously on pasting a link, no toolbar would appear. Still need to find a solution for that.

Test plan: `npm run dev && npm run test && npm run start`; in demo, paste normal links, video links, video links with extra query params etc:

- https://nodesecurity.io/advisories/755 (a normal link)
- https://www.youtube.com/watch?v=s5VpRV9yfuc (a video)
- https://youtu.be/s5VpRV9yfuc?t=178 (same video, but with a starting
  timestamp)

Closes: https://github.com/liferay/alloy-editor/issues/1109

### DIFF
--- a/src/plugins/embed.js
+++ b/src/plugins/embed.js
@@ -1,3 +1,5 @@
+import {HIGH_PRIORITY} from './priorities';
+
 /* istanbul ignore if */
 if (!CKEDITOR.plugins.get('ae_embed')) {
 	let REGEX_HTTP = /^https?/;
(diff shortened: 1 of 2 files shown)


wincent:
@julien: Not sure how to fix the unwanted toolbar pop-ups... in the meantime, I am going to try this out in the portal and see if the behavior is any different.

wincent:
One thing in testing in portal: seems that we never did auto-embeds there, so there is no change in behavior before → after this change.

jbalsas:
The problem with auto-embeds in Portal is that they rely on an external (paid) service. The expectation for the enterprise version is that everything is included. Hence, no additional services should be required.

For this reason, we forcefully remove the `ae_embed` plugin in portal in https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-editor/frontend-editor-alloyeditor-web/src/main/resources/META-INF/resources/alloyeditor.jsp#L219

wincent:
Thanks for the context @jbalsas.

julien:

PR #1113, opened 2019-02-14
Fixes #1112 | Fix image selection in IE 11

### LINKED ISSUE #1112: Fix image selection test in IE 11
There's a bug in IE 11 which prevents the toolbar from showing when an image is selected.

Open the demo in IE 11 and select an image

Current behaviour:
No toolbar are shown when the image is selected.

Expected behaviour:
A toolbar is shown when the image is selected.

julien:


src/selections/selection-test.js
```diff
@@ -73,13 +73,15 @@ let linkSelectionTest = function(payload) {
...
+	const element = selectionData.element;
+	const hasImage = !!element && !!element.findOne('img');
+	const isImage = !!element && element.getName() === 'img';
 
 	return !!(
-		selectionData.element &&
-		selectionData.element.getName() === 'img' &&
-		!selectionData.element.isReadOnly()
+		element &&
+		(hasImage || isImage)
```
wincent: Tiny nit here: on lines 83 and 84 you're relying on the truthy/falsiness of the expressions, but everywhere else you are coercing to booleans with `!!`. Personally I think The JS Way™ is probably to rely on mere truthy/falsiness, until you get to the function boundary, where you return an actual boolean.

wincent:
(Not necessary for this PR, but...) in the future we should go through all these:

```javascript
let someFunction = function() {
  /* ... */
};
```
and turn them into:

```javascript
function someFunction() {
  /* ... */
};
```
Could probably do it with a codemod.

src/selections/selection-test.js
julien: So something like this?

```javascript
const selectionData = payload.data.selectionData;
const element = selectionData.element;
const hasImage = element && element.findOne('img') !== null;
const isImage = element && element.getName() === 'img';

return !!(
  element &&
  (hasImage || isImage)
);
```

julien:
> Could probably do it with a codemod.

I did it for this file at least 👍

src/selections/selection-test.js
wincent: Yeah. In general, my thinking on coercing to bool is that I only want to do it where the type checker tells me I have to do it, in order to write "idiomatic" JS. I know we're not using a type-checker, but as a thought experiment, you can image we'd annotate this function as having return type `boolean`, which means that we'd only need one `!!` (in the expression returned by the`return` statement).

https://www.typescriptlang.org/play/#src=function%20booly(a%3A%20string%2C%20b%3A%20string)%3A%20boolean%20%7B%0D%0A%20%20return%20!!(a%20%26%26%20b)%3B%0D%0A%7D

wincent:
> I did it for this file at least 👍

Awesome. 🎉

src/selections/selection-test.js
julien: I see - thanks for the context.
Here `selectionData.element` can be `null` that's why I was using these `!!` all over the place but they're not needed

PR #1116, opened 2019-02-14
Update .prettierignore to be aware of new location of generated files

wincent:
I failed to include this when I prepared #1107. If I had run `npm run format` I would have noticed.

Test plan: run `npm run format` and see no changes.


wincent:
I'll just merge this straight away because it is utterly trivial.

Issue #1123, opened 2019-02-15
Confirm that all tests are actually running when doing `npm run test`


wincent:
I don't think all the tests are running when you do `npm run test`. Sample output:

```
Suites and tests results:

 * should throw exception when no parent or child : ok
 - Debounce :
   * debounces function execution : ok
   * calls debounced function with additional alguments : ok
   * debounces function execution with context and params : ok
   * detaches a debounced function execution : ok
   * debounces function execution for the specified delay : ok
 - Link :
   * should create a link from a selection : ok
   * should create a link from an empty selection : ok
   * should add default protocol when creating a link : ok
   * should not add default protocol if it already has one : ok
   * should not add default protocol if link is an on page bookmark : ok
   * should add mailto: when creating an email link : ok
   * should not add mailto: to an email link if it already has mailto: : ok
   * should not add mailto: to a link with an @ if after a forward slash : ok
   * should retrieve a link from a selection : ok
   * should retrieve a link from text selection : ok
   * should return null when there is text selection but not link : ok
   * should remove a passed link element : ok
   * should remove a link from a selection : ok
   * should update the URL of a passed link element : ok
   * should update the URL of a link from selection : ok
   * should update a set of attributes of a passed link element : ok
   * should update a set of attributes of a link from selection : ok
   * should remove a set of attributes from a passed link element : ok
   * should remove a set of attributes from a link from selection : ok
   * should add default protocol when updating a link : ok
   * should position the cursor before the next word if modifySelection.advance is set to true : ok
   * should position the cursor after the link if it is the last word and if modifySelection.advance is set to true : ok
   - when appendProtocol is false :
     * should not add default protocol when creating a link : ok
     * should not add default protocol when updating a link with string parameter : ok
     * should not use the default protocol when editing an existing link with object parameter : ok
 - SelectionRegion :
   - with focusing the editor after instance create :
     * should create selection from range : ok
     * should create empty selection from point : ok
     * should retrieve the caret region : ok
     * should check if the selection is empty : ok
     * should retrieve client selection regions : ok
     * should retrieve selection direction : ok
   - without focusing the editor after instance create :
     * should not throw exception on retrieving the caret region : ok
 - Table :
   * should create a 1x1 table when no parameters are specified : ok
   * should create a 1x1 table when an empty configuration is specified : ok
   * should create a 3x1 table when only the rows param is specified : ok
   * should create a 1x3 table when only the cols param is specified : ok
[...]

PR #1131, opened 2019-02-18
Fix tests broken by Zepto removal

wincent:
I broke some tests by removing Zepto in:

https://github.com/liferay/alloy-editor/pull/1087

Didn't notice though because many of our tests are silently failing; related issue that I'm currently working on for that:

We can replace these Zepto usages with classList because it is available on all of our supported platforms:

https://caniuse.com/#feat=classlist

Test plan: run tests in browser with Karma config `singleRun: false` so that I can see the failures no longer appear in the console.


julien:

PR #1135, opened 2019-02-19
Add test:debug script for investigating test issues

wincent:
A few changes in here:

- Split out one of the preprocessor patterns into two lines so that readers don't have to look up the significance of "+"; if you're curious, it is explained [here](https://github.com/isaacs/minimatch).
- Create `DEBUG` constant (set when running with the `-d` or `--debug` flags) for readability and use it in a few places.
- When `DEBUG` is set, increase log level from "info" to "debug".
- When `DEBUG` is set, set "singleRun" to `true`, which means you can interact with the tests in the browser console.
- Add a "test:debug" script to run with `--debug`; this is a shortcut for doing `npm run test -- --debug`.
- Produce coverage information for "src", not "test"; seems like this was back-to-front.

Doing this in service of:

### DIFF
--- a/karma.common.js
+++ b/karma.common.js
@@ -7,12 +7,15 @@ const alloyEditorDir = 'dist/alloy-editor/';
 
 const preprocessors = {
 	'test/**/*.html': ['html2js'],
-	'+(test|src)/**/*.js*': ['webpack'],
+	'src/**/*.js*': ['webpack'],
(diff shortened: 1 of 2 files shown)


julien:
Just started reviewing 👀 <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

PR #1144, opened 2019-02-20
Changed the height of the header in desktop

edalgrin-work:


src/styles/_general.scss
```diff
@@ -22,6 +22,12 @@
 	min-height: 100vh;
 }
 
+.min-vh-lg-100 {
+	@include media-breakpoint-up(lg) {
+		min-height: 100vh;
```
wincent: Am I reading correctly that this makes the header the full height of the viewport on large screens? Seems pretty extreme.

PR #1146, opened 2019-02-20
Test fixes

wincent:
Lots of changes here, detailed in the individual commit messages, but the high-level summary is:

- Fixed a misnamed file.
- Factored out repeated `this.skip()` calls (and incidentally, means that we no longer try to set up the editor with the dragresize_ie plugin when not on IE).
- Added a check-and-warning for bad state left behind in the DOM, and clean it up so that subsequent tests don't fail spuriously.
- Refactor before/after hooks to make them easier to use.
- Update some straggling legacy `sinon.stub()` calls that were causing failures.
- Fixed usage of arrow functions that was messing with `this` and thus causing tests to fail.
- Simplify `this.render()` API in tests.
- Remove a bunch of `React.findDOMNode` calls that weren't working and were causing tests to fail.

Surely more to come after this.

Related: https://github.com/liferay/alloy-editor/issues/1123


wincent:
Closing this one: pushed to https://github.com/liferay/alloy-editor/pull/1148 instead so that @julien can add commits.

PR #1147, opened 2019-02-20
[DO NOT MERGE] Fixes 2 UI tests

julien:


julien:
@wincent this doesn't feel right at all but I thought it would be interesting to have a look at later

julien:
@wincent closing this for now, will put this on top on #1146

PR #1148, opened 2019-02-21
Test fixes

wincent:
This replaces #1146. Pushing it here to the canonical repo so that @julien can add commits to it too.

Related #1123.


wincent:
Getting close now. Used to have screenfuls of failures. Now just this:

karma_debug_runner

Issue #1150, opened 2019-02-22
Make it possible to run with non-minified CKEDITOR build


wincent:
While working on #1148 some of the tests would end up throwing exceptions deep inside the minified CKEDITOR source. You can get the debugger to "pretty print", but it's still super hard to find your way around because all the variable names are minified, like in this sample snippet:

```
    function a(a) {
      var f = d(this);
      return f[a] || (f[a] = new b(a))
    }
    var d = function (a) {
      a = a.getPrivate && a.getPrivate() || a._ || (a._ = {
      });
      return a.events || (a.events = {
      })
    },
    b = function (a) {
      this.name = a;
      this.listeners = [
      ]
    };
    b.prototype = {
      getListenerIndex: function (a) {
        for (var b = 0, d = this.listeners; b < d.length; b++) if (d[b].fn == a) return b;
        return - 1
      }
    };
```

So, the idea would be to make a debug build that we use in `npm run test:debug` to make the debug experience better.

blzaugg:
I would want do be a able to run a non-minified distribution as well. For example, in Portal.

wincent:
@blzaugg: I think that would be doable. At the moment the portal requires CKEDITOR
and then uses the `alloy-editor-no-ckeditor-min.js` build [here](https://github.com/liferay/liferay-portal/blob/37e20bd0e38350e56a2de1aaacc507895d9e2202/modules/apps/frontend-editor/frontend-editor-alloyeditor-web/src/main/resources/META-INF/resources/resources.jsp#L38).

So to be able to use a non-minified CKEDITOR in portal two things would be required:

1. Teach webpack to make a non-minified build for us; should be relatively straightforward and similar to what is visible in this #1155 and #1156.
2. Update 
[this line in the portal repo](https://github.com/liferay/liferay-portal/blob/37e20bd0e38350e56a2de1aaacc507895d9e2202/modules/apps/frontend-editor/frontend-editor-alloyeditor-web/src/main/resources/META-INF/resources/resources.jsp#L36) to conditionally require the minified build instead when debugging. I'm new to that code base so I don't know exactly how you'd do that; might be by setting a property in the `bundles/portal-ext.properties` file alongside your portal checkout.

I guess if you were running with non-minified CKEDITOR, you'd probably also want to run non-minified AlloyEditor at the same time too.

I created an internal ticket to track this: https://issues.liferay.com/browse/IFI-488

PR #1154, opened 2019-02-22
Fix embed tests

wincent:
Alas, I have to reintroduce the hack that I removed in c7f77084e7a5631e4814c92bf1. Notice though that I am not applying it in a blanket fashion everywhere -- that just makes other failures pop up elsewhere -- but instead am targeting it to the two places that need it. There are only two other failing tests in Firefox visible with `npm run test:debug` right now. If they require a similar fix I'll extract out the hack.

This is some really baffling behavior. To summarize:

- Tests only fail on Firefox, and only if a test has run before.
- In fact, it is not even necessary to run a test before to trigger the failure: just adding a "div" with a "contenteditable" attribute to the DOM is enough (removing it makes no difference).
- As I was single-stepping through the debugger, I saw CKEDITOR incorrectly attaching itself to the first contenteditable that it found, even if it did not have the "#editable" ID that we are passing in out call to `CKEDITOR.inline`, and even if we pass the actual DOM object instead of passing the ID name.
- When you set the selection using this hack, the problem magically goes away.

I'd still like to try this with a non-debug build of CKEDITOR -- see https://github.com/liferay/alloy-editor/issues/1150 -- so I can see what the actual error that is being thrown is, but for now it's just an inscrutable 'TypeError: "B is undefined"'.

Bonus: use Sinon sandboxes for better tear-down, and make tear-down in general more resilient.

### DIFF
--- a/test/lib/utils.js
+++ b/test/lib/utils.js
@@ -197,8 +197,11 @@ if (!window.Utils) {
 				this.nativeEditor = null;
 			}
 
-			this._editable.parentNode.removeChild(this._editable);
-			this._editable = null;
+			if (this._editable) {
+				this._editable.parentNode.removeChild(this._editable);
+				this._editable = null;
+			}
+
 			Utils.removeContainer.call(this);
 			fixture.cleanup();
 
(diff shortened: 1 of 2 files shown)


julien:
Just started reviewing :)

Issue #1124, opened 2019-02-15
Automate and Improve CKEditor update process


jbalsas:
The current process of updating our bundled version of CKEditor is currently manual and involves a series of [custom manual steps](https://github.com/liferay/alloy-editor/blob/master/CONTRIBUTING.md#updating-ckeditor) as describe in the [CONTRIBUTING guidelines](https://github.com/liferay/alloy-editor/blob/master/CONTRIBUTING.md)

CKEditor 4.x is currently being released as the [ckeditor npm package](https://www.npmjs.com/package/ckeditor), so that could be an option. If some of their [releases](https://github.com/ckeditor/ckeditor-releases/tree/standard/4.11.x#releases-code) matches our needs, we could simply pick from there.

A more custom version could be achieved by using the proper [release branch](https://github.com/ckeditor/ckeditor-dev/tree/release/4.11.x) and running the build process with our predefined `config.json` settings.

Doing a quick cost-benefit analysis, I think we could just pick one of the official releases probably taking a bit of a hit on bundle size so we can avoid a overly complex update process.

julien:
Seems like a good idea 👍

blzaugg:
I remember looking into trying to use their [online builder](https://ckeditor.com/cke4/builder) to programmatically (using `build-config.js`) build a custom build that fit Portal's needs. But I couldn't identify an established API to do that. It seemed any hacked solution could break at any time in the future.

jbalsas:
The best way is to simply use their own packaging strategy -> https://github.com/ckeditor/ckeditor-dev/tree/major/dev/builder

PR #1155, opened 2019-02-22
Automate CKEditor build process

wincent:
Three things here:

## (i) Replace downloaded CKEDITOR build files with locally-built versions

These correspond to the v4.11.2 release that we are currently using.

## (ii) Add debug (unminified) CKEDITOR build at lib/ckeditor-debug/

This is a prerequisite for:

https://github.com/liferay/alloy-editor/issues/1150

Not using it yet in this PR, just preparing to. (See #1156 for the follow-up.)

This build also corresponds to the v4.11.2 version that we are currently using.

## (iii) Add a script for performing a build

Invoked like this:

    scripts/build/build-ckeditor.sh 4.11.2

Update CONTRIBUTING.md to reflect the existence of the tool, and made sure that dirty status in the ckeditor-dev submodule won't clutter up `git status` output (it's normal to dirty the submodule because we copy our own "build-config.js" into it).

Test plan: run the script and see that the build worked and the only changes made to the files were timestamp updates.

Closes: https://github.com/liferay/alloy-editor/issues/1124


blzaugg:
Is there a reason we're adding the CKEditor source to our project, rather than downloading it? With one built `ckeditor.js` file it seemed "okay" but now this seem unnecessarily heavy.

wincent:
@blzaugg: You either end up downloading it with `curl` or with `git clone`/`git fetch`, so this seemed like the simplest way. With the change proposed here, once you've cloned the repo and run `npm install`, you have everything necessary to build and debug the project.

Additionally, Git is very good at compressing source code, and you get the added benefit that you can easily see what changed between versions by invoking `git diff`, which can be another useful debugging aid.

julien:
Just started reviewing :)

blzaugg:
I'm more thinking about when we update CKEditor. We'd need to update all those files, rather than just a version reference.

Also, would those file updates automated, or is it a manual step?

wincent:
> Also, would those file updates automated, or is it a manual step?

In this PR you just run `scripts/build/build-ckeditor.sh $VERSION`. In the case of [liferay-ckeditor](https://github.com/liferay/liferay-ckeditor) I don't think the same approach applies, because you need to not only apply the patches on top (which may be non-trivial, given the [large amount of changes since v4.6.0](https://github.com/ckeditor/ckeditor-dev/compare/4.6.0...4.11.2)), but actually evaluate to see whether they're still even required; some bugs may have been fixed upstream for example, or otherwise been rendered invalid or redundant or inapplicable etc. I created [this ticket](https://issues.liferay.com/browse/LPS-91153) to track updating liferay-ckeditor to v4.11.2, which is the one used in this repo; we'll take a look at that this sprint.

blzaugg:
Thanks. I see it's documented here: https://github.com/liferay/alloy-editor/blob/master/CONTRIBUTING.md#updating-ckeditor

PR #1157, opened 2019-02-22
Fix button-embed-edit tests

wincent:
With this we now have a 100% clean `npm run test:debug` run (in browser, in Firefox); that's around 350 tests.


wincent:
I'm just going to merge this one because it is pretty trivial and I want to build something on top of it.

wincent:
For reference this is the next step that I wanted to do: https://github.com/liferay/alloy-editor/pull/1158

PR #1158, opened 2019-02-25
Streamline and fool-proof-ize the clean up of Sinon stubs

wincent:
Sinon has a "sandbox" feature that allows you to create stubs anywhere and then clean them up with a single call to `sandbox.restore()`. I just learned that as of v5, the `sinon` object is itself a sandbox, which means that we can:

1. Remove the places where we manually set up a new sandbox and just call `sinon.stub()` etc directly.
2. Add a `sinon.restore()` call in a global "afterHook" and be sure that clean-up is never forgotten.
3. Remove existing scattered calls to `restore()`.

See:

    https://sinonjs.org/releases/latest/sandbox/

Test plan: `npm run build && npm run test:debug` and see all tests pass.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -41,6 +41,7 @@ module.exports = function() {
 		'test/lib/globals.js',
 		'test/lib/utils.js',
 		'test/lib/karma-overrides.js',
+		'test/lib/mocha.js',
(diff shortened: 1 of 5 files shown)


julien:
Just started reviewing :)

PR #1159, opened 2019-02-25
Update to react/react-dom 16.8.3

wincent:
As well as adding "^" to the version range, which should be totally safe with React because they follow semver very rigorously.

Test plan: sanity check by running `npm run build && npm run test:debug`


julien:
Just started reviewing :)

PR #1162, opened 2019-02-27
Fixes #1160 Add transitionend event to enable pointerEvents

### LINKED ISSUE #1160: IE11 double-clicking word result in click icons in toolbar (underline/bolded..)
**Do you want to request a *feature* or report a *bug*?**
BUG

**What is the current behavior?**
Using IE11 when double clicking a word, the icons in the toolbar are accidentally clicked.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
https://issues.liferay.com/browse/LPS-91194
It is hard to see in the AE Demo and noticeable in Liferay.
1. Use AE demo in IE11
2. Double click a word so the toolbar appears.

Result: A icon is clicked, in this case the font style is selected.

**What is the expected behavior?**
No icon in the toolbar should be clicked when double clicking a word.

IE11, 2.0.0-beta.1

fortunatomaldonado:
https://github.com/liferay/alloy-editor/issues/1160
https://issues.liferay.com/browse/LPS-91194

I included `pointerEvents: 'none'` to the style to prevent any unwanted button selections. Then applied a `transitionend` event listener. Only when the transition has completed will the pointerEvents be removed and allow the user to interact with the toolbar.

Thank you.

### DIFF
--- a/src/components/base/widget-position.js
+++ b/src/components/base/widget-position.js
@@ -319,6 +319,7 @@ export default WrappedComponent =>
 				left: startPoint[0] + 'px',
 				top: startPoint[1] + 'px',
 				opacity: 0,
+				pointerEvents: 'none',
 			});
 
 			domElement.removeClass('alloy-editor-invisible');
@@ -332,6 +333,14 @@ export default WrappedComponent =>
 					opacity: 1,
 				});
 			});
+
+			if(domElement.$){
+				domElement.$.addEventListener('transitionend', () =>{
+					domElement.setStyles({
+						pointerEvents: '',
+					});
+				})
+			}
 		}
 
 		/**
--- a/src/components/base/widget-position.js
+++ b/src/components/base/widget-position.js
@@ -334,8 +334,8 @@ export default WrappedComponent =>
 				});
 			});
 
-			if(domElement.$){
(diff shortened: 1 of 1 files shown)


julien:
Just started reviewing :)

src/components/base/widget-position.js
```diff
@@ -332,6 +333,14 @@ export default WrappedComponent =>
 					opacity: 1,
 				});
 			});
+
+			if(domElement.$){
```
wincent: nit: space after the `if `

And also after the `=>` below.

You should probably run `npm run format`.

julien:
@fortunatomaldonado I tested in IE11 and it seems to work fine, however take into account @wincent 's comment concerning formatting.
Thanks!

src/components/base/widget-position.js
fortunatomaldonado: Hello @wincent,

Thank you, I made the format changes. I was unable to run npm run format though.

fortunatomaldonado:
Hello @julien,

Sounds good. I added the necessary format changes. Let me know if there is anything else I should do.
Thank you.

julien:
Just started reviewing :)

PR #1166, opened 2019-03-01
Make it convenient to check out website as worktree

wincent:
So that you can run `git worktree add website origin/website` (or similar) and have a handy checkout of the website branch in a way that doesn't interfere with work on the master branch.


wincent:
This is pretty trivial so going to merge it.

PR #1169, opened 2019-03-05
Update to eslint-config-liferay v3.0.0

wincent:
No major changes, but it does allow us to turn off a suppression we we no longer need ("sort-keys").


julien:
Just started reviewing :)

PR #1185, opened 2019-03-20
Fix getImageAlignment for IE11

julien:


scripts/build/gulp/demo/index.html
```diff
@@ -125,5 +125,8 @@ <h3>Full toolbar demo</h3>
                 - Frederico Knabben. All rights reserved.
             </p>
         </div>
+
+				<script src="//unpkg.com/svg4everybody@2.1.9/dist/svg4everybody.js"></script>
+				<script>svg4everybody();</script>
```
wincent: I know this file should probably use tabs, but it seems the rest of it is currently using spaces, so the indentation looks a bit whack:

Fix_getImageAlignment_for_IE11_by_julien_·_Pull_Request__1185_·_liferay_alloy-editor

julien: You're right, I'll fix that right now

Issue #1189, opened 2019-03-27
AlloyEditor.React is undefined after upgrading 


lpsBetty:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
I added custom buttons with `var React = AlloyEditor.React`, but `AlloyEditor.React` is not available anymore. As described [here](https://alloyeditor.com/docs/develop/create_buttons.html).

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
Install the latest version of Alloy-editor (2.0.0-beta.3). Try to access `AlloyEditor.React` in the console - it will be undefined.

**What is the expected behavior?**
I would like to access React and create Buttons.

**Which versions of alloy-editor, and which browser / OS are affected by this issue? Did this work in previous versions?**
Chrome on macOS,  Alloy-editor 2.0.0-beta.3
It worked with Alloy-editor 1.5.19

lpsBetty:
@julien i still got 1 error now:

```
Uncaught TypeError: React.createClass is not a function
```

Has the method name been changed? I create the buttons like described in your [documentation](https://alloyeditor.com/docs/develop/create_buttons.html).

julien:
Hi @lpsBetty,

Yes that error was expected.

First of all, it's tue that we need to update our documentation, apologies for not having done so.

`React.createClass` was actually removed a [while back](https://reactjs.org/blog/2017/09/26/react-v16.0.html) from `React` itself and now the way to create `React` components is done by extending the `Component` class like so:

```javascript
import React from 'react';

class MyComponent extends React.Component {

}
```

You can have a look at some existing [**button** components](https://github.com/liferay/alloy-editor/tree/master/src/components/buttons) to have a more detailed example.

Let me know if that helps.

PR #1190, opened 2019-03-27
fix: Adds AlloyEditor.React

julien:


julien:
@wincent this fixes #1189 as discussed previously

scripts/build/with-react.js
```diff
@@ -0,0 +1,13 @@
+import '../../lib/ckeditor/ckeditor';
```
wincent: Seems like we shouldn't have this line here, because it will pull ckeditor into the `noCkeditor` build, won't it?

scripts/build/with-react.js
```diff
@@ -0,0 +1,13 @@
...
+import * as AlloyEditor from '../../src/adapter/main';
+import React from 'react';
+
+if (typeof window !== 'undefined') {
+	window.CKEDITOR.disableAutoInline = true;
+	window.AlloyEditor = AlloyEditor;
+	window.AlloyEditor.React = React;
+}
+
+export {AlloyEditor, React};
```
wincent: Do we need to export `React` here?

scripts/build/with-react.js
julien: Right, this shouldn't be there

scripts/build/with-react.js
julien: We don't ... but we had fun testing this thanks!

scripts/build/index.js
```diff
@@ -4,8 +4,7 @@ import * as AlloyEditor from '../../src/adapter/main';
 
 if (typeof window !== 'undefined') {
 	window.CKEDITOR.disableAutoInline = true;
-	window.AlloyEditor = AlloyEditor;
 }
 
 export {AlloyEditor};
-export default AlloyEditor;
+
```
wincent: Unless GitHub is misrendering this, looks like an excess line.

PR #1193, opened 2019-03-28
Fix pointerEvents on firefox (and IE)

julien:


src/components/base/widget-position.js
```diff
@@ -315,6 +315,12 @@ export default WrappedComponent =>
 				ReactDOM.findDOMNode(this)
 			);
 
+			domElement.on('transitionend', () => {
+				domElement.setStyles({
+					pointerEvents: '',
+				});
+			});
+
```
wincent: One thing to try before merging this solution: we might actually be able to avoid the whole "transitionend" thing if we change the `_animate()` implementation. I see it if doing an `if (window.requestAnimationFrame)` check and calling the `callback()` immediately if it is not present. I wonder if that explain the original behavior being different in IE (although "caniuse" *does* say that IE 11 has `requestAnimationFrame`). Changing that `else` to `setTimeout(callback, 0)` seems like the correct thing to do, even if it doesn't obviate the need for the "transitionend" trick.

In any case, I am happy to merge this as-is so I am hitting the "approve" button.

konradoboza:
Hi,

I stumbled upon `pointer-events: false` also in 1.5.21. Could it be ported and released against 1.5.25?

Thanks!

wincent:
@konradoboza: I believe this change is already in the 1.5 series — you can see it here, in v1.5.24, for instance:

https://github.com/liferay/alloy-editor/blob/v1.5.24/src/ui/react/src/components/base/widget-position.js#L196-L201

wincent:
Ah, I see. I only searched for where the code was added (in `v1.5.19-18-g502bcd76` — ie. 18 commits after v1.5.19 release — according to `git describe`), not to see if it was still there.

You're welcome to try your hand at a PR like you offered earlier but we need to tread carefully here and make sure this is very well tested in multiple browsers, because what we have here is patches-on-patches-on-patches and it is very easy to break things in one place when we try to fix them in others. What do you think @julien?

konradoboza:
Actually, after a deeper investigation, I figured out that this change doesn't fix my problem. So if this is much effort I guess we can drop it for 1.5.x. Up to you guys. :)

julien:
Ok thanks @konradoboza.
In anycayse, we're always happy to have contributors submit PR's.

PR #1195, opened 2019-03-29
fix: Use setTimeout instead of transitionend

julien:


src/components/base/widget-position.js
```diff
@@ -338,6 +332,12 @@ export default WrappedComponent =>
 					top: endPoint[1] + 'px',
 					opacity: 1,
 				});
+
+				setTimeout(() => {
+					domElement.setStyles({
+						pointerEvents: '',
+					});
+				}, 100);
```
wincent: As discussed in Slack, would be best if this were 150 and had a comment so that people grepping can find the related parts; eg.

```js
// 150ms to match transition-duration for .ae-toolbar-transition:
```

PR #1199, opened 2019-04-01
fix: missing PropTypes on AlloyEditor.React (#1200)

wincent:


julien:
Just started reviewing :)

Issue #1200, opened 2019-04-01
AlloyEditor.React.PropTypes is not defined


wincent:
We still use `PropTypes` internally, even though it is no longer bundled with React. Just had a report from @ricardocousolr about a failure on current master stemming from trying to access `React.AlloyEditor.PropTypes`. Given that we still use it, we should expose it in the same way as before as people may be relying on it.

Draft fix is already up at #1199 but I will clean it up and update.

PR #1201, opened 2019-04-01
fix: make React available on AlloyEditor in "all" build variant (#1189, #1200)

wincent:
Noticed while preparing the fix for #1200 (#1199) that we hadn't applied the fix for #1189 to the "all" build variant, so this commit does that.

Related: https://github.com/liferay/alloy-editor/issues/1200
Closes: https://github.com/liferay/alloy-editor/issues/1189


julien:
Just started reviewing :)

julien:
Just started reviewing :)

Issue #1197, opened 2019-03-29
Website build instructions are out of date


wincent:
https://alloyeditor.com/docs/develop/how-to-build-it.html says that you have to run `gulp build` to build, but that's not true, at least not on v2.0.x and up (now you do `yarn run build`). We should also note the minimum Node version required. Off the top of my head I don't now exactly what that is. I've been using a v11 release, but I think it should probably work with v10, most likely with v9 too, and perhaps even with v8. But we should test to confirm.

wincent:
Actually, I realize this particular one was already mentioned in https://github.com/liferay/alloy-editor/issues/1161 — rather than closing this as a dupe I'll add a back-link to this from there.

PR #1203, opened 2019-04-01
doc: Update build instructions for v2.x (#1197)

wincent:
Note that we have to keep the old 1.x instructions around as well, because we continue to make bug-fix releases on that branch.

Closes: https://github.com/liferay/alloy-editor/issues/1197


wincent:
Looks like this:

build

julien:
Just started reviewing :)

PR #1206, opened 2019-04-01
fix: run Prettier

wincent:
Looks like `yarn format` hasn't been run for while, so these files were causing `yarn checkFormat` to fail.


julien:
Just started reviewing :)

PR #1210, opened 2019-04-01
doc: update static build files

wincent:
These got updated as a result of running `yarn start` (funnily, `yarn build` doesn't touch these).


wincent:
Tested the demo still works:

demo

julien:
Just started reviewing :)

PR #1211, opened 2019-04-01
doc: document deployment process

wincent:
And remove now-unused `gh-pages` package.

This shows the steps we took to publish an update to the website just now. This will change soon because we'll have to move off of WeDeploy, but may as well keep the README.md current in the meantime.


julien:
Just started reviewing :)

PR #1214, opened 2019-04-01
fix: Include version in AlloyEditor

julien:
@wincent we didn't really discuss of how we should do this, but I thought I'd send this as a draft.


scripts/build/with-react.js
```diff
@@ -7,6 +7,7 @@ if (typeof window !== 'undefined') {
 	window.AlloyEditor = AlloyEditor;
 	AlloyEditor.React = React;
 	AlloyEditor.React.PropTypes = PropTypes;
+	AlloyEditor.version = require('../../package.json').version;
```
wincent: This is good but I'd go with `VERSION` to make it clear that this is a "constant".

PR #1216, opened 2019-04-01
fix: Add VERSION constant to AlloyEditor

### LINKED ISSUE #951: How to get the AlloyEditor version inside a Liferay 7.1 instance?
On a Liferay 7.1 installation, is there a way to get the AlloyEditor version installed?

JavaScript key `AlloyEditor.version` does not exist: any other methods to get it?

Thank you.

julien:
Closes #951

### DIFF
--- a/scripts/build/version.js
+++ b/scripts/build/version.js
@@ -1,0 +1,9 @@
+'use strict';
+const version = require('../../package.json').version;
+
+module.exports = () => {
+	return {
+		code: `module.exports = ${JSON.stringify(version)};`
+	};
+};
+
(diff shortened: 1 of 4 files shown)


scripts/build/version.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+const version = require('../../package.json').version;
+
+module.exports = () => {
+	return {
+		code: `module.exports = ${JSON.stringify(version)};`
+	};
+};
+
```
wincent: One possible gotcha here is that last time I used val-loader I wrote:

https://github.com/liferay/alloy-editor/blob/11dc22f6666e4735c3ba499478f7a3ae40214c0d/test/index.js#L22-L23

I can't remember now whether that is a test-specific thing or not, and it is probably harmless to use ES6 here as we'll always be building on a machine with modern Node.

Issue #1212, opened 2019-04-01
Migrate website away from WeDeploy


wincent:
Right now I can't remember whether there were technical reasons why we can't host the website on GitHub pages — maybe there are, maybe there aren't — but just going to document the need to migrate away here within the next few weeks (WeDeploy will be [shutting down on 22 April 2019](https://wedeploy.com/blog/discontinuing-wedeploy/)).

If we move to GitHub pages, we probably have to enable it in the settings (sample screenshot from a repo of mine):

![Options](https://user-images.githubusercontent.com/7074/55333352-69015280-5497-11e9-8430-033204e49e6f.png)

In theory, we should be able to continue using the custom domain, [according to these docs](https://help.github.com/en/articles/using-a-custom-domain-with-github-pages). Failing that, the site would be available at https://liferay.github.io/alloy-editor/.

cc @jbalsas @julien

blzaugg:
>  technical reasons why we can't host the website on GitHub pages — maybe there are, maybe there aren't 

I assumed it was to "drink your own champagne".

wincent:
https://github.com/liferay/alloy-editor/pull/1233 is the last change required here. Going to close this now because it won't autoclose (only stuff merged to "master" will autoclose issues, and that PR is going to the "website" branch).

PR #1217, opened 2019-04-01
feat: Add deploy.sh script to publish to GitHub pages (#1212)

wincent:
This commit adds a shell script that takes the contents of the "public/" directory and creates a new commit from them on the "gh-pages" branch.

If you run `./deploy.sh`, you should see something like the following (the dots are progress indicators, printed one per file as we go seeing as we are forking a lot and it isn't super fast; it takes about 11 seconds on my machine to run the script):

```
Working from top level: /Users/greghurrell/code/alloy-editor/website
Clearing index
Adding public/ files to index
.......................................................................
.......................................................................
.......................................................................
.......................................................................
.......................................................................
.......................................................................
...................................
Created tree object 462d4368bb55c356147afe3c4e001ce2434dbdb3
Creating commit object on gh-pages branch
commit 6f4e2d29ae46ef339571877f7e42435883a079e5 (gh-pages)
Author: Greg Hurrell <greg.hurrell@liferay.com>
Date:   Mon Apr 1 18:38:44 2019 +0200

    Autogenerated public/ content from f606edab
Restoring index to prior state
```

Caveats: because `public/` is in .gitignore, we can't produce this from the `HEAD` and instead just take whatever is on the disk. Hopefully that corresponds to what you'd get from running `yarn build` on HEAD, but there are no guarantees.

Likewise, it only creates the commit on the "gh-pages" branch -- it doesn't actually push it -- so that you can review it.

Test plan: run the command as above and see it produce this commit on the branch:

https://github.com/liferay/alloy-editor/commit/6f4e2d29ae46ef339571877f7e42435883a079e5

Related: https://github.com/liferay/alloy-editor/issues/1212


deploy.sh
```diff
@@ -0,0 +1,54 @@
...
+  HASH=$(cat $SRC | git hash-object -w --stdin --path .)
+  DEST=$(/bin/echo -n $SRC | sed -e 's;^public/\(.*\);\1;')
+
+  if [ -x "$SRC" ]; then
+    MODE=100755
+  else
+    MODE=100644
+  fi
+
+  /bin/echo -n "$MODE blob $HASH	$DEST" | git update-index --index-info
```
wincent: The big gap in there is a literal tab, in case you are wondering.

deploy.sh
```diff
@@ -0,0 +1,54 @@
...
+ABBREV=$(git rev-parse --short HEAD)
+COMMIT=$(git commit-tree -m "Autogenerated public/ content from $ABBREV" -p gh-pages "$TREE")
+git branch -f gh-pages "$COMMIT"
+git log -1 gh-pages
+
+echo "Restoring index to prior state"
+git reset HEAD
+
+echo
+echo "Now run 'git push upstream gh-pages'"
```
wincent: Obviously not everybody uses "upstream" as a branch name, but hopefully the required action is clear anyway. If people think it's necessary we can try to infer the right remote name from the output of `git remote -v`.

julien:
Just started reviewing :)

PR #1218, opened 2019-04-02
Fix toolbar location for images

julien:


wincent:
Next time let's use "fix: toolbar location for images" (in case we decide to extract the metadata in the future in our changelogs, it would be good to get in the habit; although obviously we can trivially add a check for PR titles starting with "Fix ..." as well).

Issue #1220, opened 2019-04-04
v2.x branch uses an old version of clay-css dependency


wincent:
We noticed in the yarn.lock inside liferay-portal that we are using an old version of clay-css here (2.5.1). Portal is on a much newer 2.11.1, so we should check and see if we can move to that without breaking anything.

PR #1221, opened 2019-04-04
chore: Update to latest clay-css (#1220)

wincent:
We're using clay-css for the spritesheet, but we had it pinned to an old version. In liferay-portal, we're using clay-css 2.11.1 right now, and there's no reason why we shouldn't just use the same here. So I'm specifying a looser range. Note that the latest is actually 2.12.0, but I only ask for "^2.11.1" because I want `yarn` in the portal to resolve clay-css to a single version.

Test plan: `yarn dev && yarn start`

testing

Closes: https://github.com/liferay/alloy-editor/issues/1220

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
   },
   "homepage": "http://alloyeditor.com",
   "dependencies": {
-    "clay-css": "2.5.1",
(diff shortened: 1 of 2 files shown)


wincent:
testing

Issue #1198, opened 2019-03-29
Using Ctrl + a in Firefox, the toolbar is centered on editor and not selected text


fortunatomaldonado:
**Do you want to request a *feature* or report a *bug*?**
Bug
**What is the current behavior?**
Using Ctrl + a in Firefox, the toolbar is centered on editor and not selected text
**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
1. Go to AE Demo in Firefox
2. Clear all the content in the content editor
3. Add any text "This is a sentence"
4. Click on text and use Ctrl + a to highlight all the text

Result: The toolbar will appear centered on the editor and not the text.

**What is the expected behavior?**
The expected result would be the toolbar be centered on the text
**Which versions of alloy-editor, and which browser / OS are affected by this issue? Did this work in previous versions?**
Master, Firefox
https://issues.liferay.com/browse/LPS-92977

julien:
@fortunatomaldonado I had a look at https://issues.liferay.com/browse/PTR-747, and it looks like fixing this issue won't actually fix the other problem because the "toolbar" moves all the time, I'll still give it a shot though

PR #1223, opened 2019-04-09
fix: fix selection-region for Firefox

julien:


julien:
@wincent the issue is #1198, however this is probably not the correct fix so I'm closing this for now.

Issue #1224, opened 2019-04-15
Drag Handles Disappear During Resizing on IE11


gregory-bretall:
**Do you want to request a *feature* or report a *bug*?**
Report a bug

**What is the current behavior?**
While resizing an image on IE11, if your cursor leaves the bounds of the image while resizing, the drag handles will disappear.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
Not currently reproducible in standalone AlloyEditor from what I can tell since the only build steps for AlloyEditor I could find locally do not transpile ES6, making it not work on IE11.

Steps for working environments should be as follows:
1. Add an image to AlloyEditor on IE11.
2. Resize the image using the drag handles.
3. Drag back and forth so that your cursor leaves the bounds of the image

Expected Results: When the cursor leaves the bounds of the image, the drag handles remain on the screen.
Actual Results: When the cursor leaves the bounds of the image, the drag handles disappear, even while still resizing.

**What is the expected behavior?**
While resizing, the drag handles should always appear on screen, regardless of cursor position.

**Which versions of alloy-editor, and which browser / OS are affected by this issue? Did this work in previous versions?**
This affects IE11 on AlloyEditor Master and 2.x as well as 1.x.

julien:
Hi @gregory-bretall,

Thanks for reporting this issue.
We will look into it.

Also, just so you know, the alloy-editor demo does work in IE11.

You can use:

`yarn start`

To start a local development server (it will use port 9000)

And

`yarn build`

To build a production ready version.

You can see this in the following .gif image

![2019-04-16-alloyeditor-demo-ie11](https://user-images.githubusercontent.com/5572/56192304-5c036800-602e-11e9-915a-56240aa99095.gif)

Thanks

julien:
Closed in:
master by [2a47932e57a8e8fda6cd33629fd4c5e5b0a89c75](https://github.com/liferay/alloy-editor/commit/2a47932e57a8e8fda6cd33629fd4c5e5b0a89c75) 
and in 1.x by [f0fbcf9f1f3789acb340512b8c720c33c29ce80c](https://github.com/julien/alloy-editor/commit/f0fbcf9f1f3789acb340512b8c720c33c29ce80c)

PR #1225, opened 2019-04-15
#1224 cke_image_resizing class is added to containing parent span for…

gregory-bretall:
…s image resizer, change selector so drag handles don't disappear while resizing

Fix for https://github.com/liferay/alloy-editor/issues/1224

Can't reproduce this issue in standalone AlloyEditor on IE11 because the only build steps I found for building AlloyEditor did not transpile ES6 into ES5 for some reason, making AlloyEditor in general not work correctly on IE11. I did, however, include steps in the case that you are able to get AlloyEditor to properly work on IE11.

Please let me know if there are any issues with the procedure here.

Related LPS: https://issues.liferay.com/browse/LPS-94092

### DIFF
--- a/src/plugins/dragresize_ie11.js
+++ b/src/plugins/dragresize_ie11.js
@@ -79,7 +79,7 @@
 					'right:auto;' +
 					'}' +
 					'.cke_widget_wrapper:hover .cke_image_resizer,' +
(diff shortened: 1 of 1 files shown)


julien:
Just started reviewing :)

Issue #1227, opened 2019-04-17
Live website demo is not working in IE11


blzaugg:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
The live demo on the website does not work in IE11, and errors in the browser console.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
1. Open IE11 to https://www.alloyeditor.com/
2. Scroll down to inline editor.
3. Attempt to edit content or interact with the editor

**What is the expected behavior?**
Should be able to edit content and interact with the editor.

**Which versions of alloy-editor, and which browser / OS are affected by this issue? Did this work in previous versions?**
Internet Explorer 11 on Windows 7

blzaugg:
![AlloyEditor com in IE11](https://user-images.githubusercontent.com/1626739/56315798-86782280-610d-11e9-9cf8-0302f1315467.gif)

wincent:
Thanks for catching this @blzaugg. Looks like we need to add a polyfill. The website is just including "alloy-editor/alloy-editor-all-min.js" and expecting everything to be rosy but evidently that's not enough.

wincent:
Ok, fix is [now deployed to gh-pages branch](https://github.com/liferay/alloy-editor/commit/caf88fe0b490d9908e3323009f57ea716c1c3ef0). Unfortunately you can't see it on [the live site](https://www.alloyeditor.com/) and in fact the live site is down right now due to a misconfigured CNAME. I've opened an IT ticket to get that fixed.

PR #1230, opened 2019-04-22
fix: make demo less broken on IE (#1227)

wincent:
Fixes two issues:

- Lack of polyfills that were cause the demo to blow up on IE 11; note that "Gatsby only automatically add polyfills for your code (and not for your project's dependencies in node_modules) at the moment." - that is why we have to manually add the polyfill rather than relying on "babel-preset-gatsby" and a ".babelrc" which sets "targets".
  - https://www.gatsbyjs.org/docs/browser-support/#polyfills
- The "browser-monads" package -- which enables you to access `window` and `document` without checking they exist, was pulling in "nothing-mock" which uses `Proxy` without first checking for its existence (how ironic); seeing as it can't be polyfilled, this was causing IE to throw an error.
  - https://www.npmjs.com/package/browser-monads
  - https://www.npmjs.com/package/nothing-mock

Test plan: `yarn build && yarn serve`, then hit http://10.0.2.2:9000/ from IE 11 running in a VirtualBox machine. No more errors and the demo kind of works. Still need an svg4everybody polyfill to make the icons show up, but I'll deal with that in another commit, which is why I am labeling this one as "Related:" and not "Closes":.

Related: https://github.com/liferay/alloy-editor/issues/1227


wincent:
@eduardoallegrini: We might have this same kind of issue on other sites that use the boilerplate; I haven't checked yet though.

blzaugg:
Should polyfils be applied at the end project level (AlloyEditor.com), or the Library level (Alloy Editor)?

blzaugg:
+1 for "less broken"

wincent:
> Should polyfills be applied at the end project level (AlloyEditor.com), or the Library level (Alloy Editor)?

I'd say library level because it provides flexibility: if your user base isn't on legacy browsers, you skip the polyfill, and if it is, you include it. This is effectively how we use it in Liferay DXP (ie. we deploy the library, and we make sure that old browsers get served a polyfill).

julien:
Just started reviewing :)

blzaugg:
Can we document (in Alloy Editor) what's "recommended", in regard to polyfills, for use in projects (AlloyEditor.com, Portal, etc.)?

Also, how can we claim "Works on IE9+", when the polyfills, are not included?

jbalsas:
Many of the polyfills are needed for the site, not for the editor, but yeah, we should document which ones like `svg4everybody` are actually necessary in IE.

> Also, how can we claim "Works on IE9+", when the polyfills, are not included?

Let's not. 2.x branch should be only IE11. We can certainly update the copy in the site

wincent:
> Let's not. 2.x branch should be only IE11.

That's right. For some context, we made the explicit decision to have AlloyEditor support the same versions as declared in the DXP compatibility matrix (which means IE11 is supported). 7.0 mentions "limited support" for older IE, but I think it's best not to muddy the waters by mentioning partial/unofficial support for IE9 on one of the branches. I'll update the website.

Issue #1228, opened 2019-04-19
Advanced Editing Demo fails with console errors in IE11


blzaugg:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Demo doesn't work and console errors in browser
```
SCRIPT1002: Syntax error
File: alloy-editor-all.js, Line: 36086, Column: 21
SCRIPT5009: 'AlloyEditor' is undefined
File: localhost:9000, Line: 109, Column: 13
```

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
1. Following the steps from [Build it](https://www.alloyeditor.com/docs/develop/how-to-build-it.html) 
```
$ yarn install
$ yarn build
$ yarn start
```
2. Open http://localhost:9000/ in IE 11

**What is the expected behavior?**
Demo works with no console errors.

**Which versions of alloy-editor, and which browser / OS are affected by this issue? Did this work in previous versions?**
Windows 7 Internet Explorer 11

blzaugg:
![Advanced Editing Demo IE 11](https://user-images.githubusercontent.com/1626739/56399014-8a38a180-6200-11e9-84af-9be72ec57b9a.gif)

blzaugg:
The syntax error is from an arrow function that isn't transpiled to ES5:
```js
module.exports = () => {
	return {
		code: `module.exports = ${JSON.stringify(version)};`,
	};
};
```

wincent:
Thanks for reporting this @blzaugg!

> The syntax error is from an arrow function that isn't transpiled to ES5:

I'm not sure if that's the function though; that file should only be evaluated as part of the build process, and should never wind up in the browser. We'll dig into it further though.

wincent:
I'm afraid I can't repro this on IE 11 @blzaugg. Any ideas what might be needed to trigger? (I'm using IE 11 in a VirtualBox VM, hitting the webpack dev server that you get with `yarn start` on port 9000.

![Screen Shot 2019-04-23 at 16 53 42](https://user-images.githubusercontent.com/7074/56591209-95b50f80-65e8-11e9-861d-ec4309fa4549.png)

gregory-bretall:
@wincent I was able to reproduce this on the most up to date master branch from today of alloy-editor, just by using yarn start. Not quite sure what the difference here is unless it's testing IE11 on Windows 10 potentially?

wincent:
[...]

PR #1236, opened 2019-04-23
fix: console errors in Windows-based build (#1228)

wincent:
It turns out that the "test" property in the Webpack build needs to be in a different format on Windows, otherwise it never matches, which means that our file at "scripts/build/version.js" doesn't pass through the val-loader, and in turn means that we produce untransformed output that IE 11 will choke on (an arrow function, and even if we fixed it, it would probably choke on the backticks).

You can make the build work on Windows by passing `path.resolve('scripts/build/version.js')` instead; this produces an absolute path with the appropriate Windows-style backslashes. But then the build doesn't work on macOS. If you try to make it a relative path (`path.relative('', path.resolve(...))`) then Webpack rejects it.

So, the solution is to make a RegExp with the backslashes (if present) escaped.

Test plan: build on macOS and Windows (with `yarn build`, `yarn start`) and confirm console errors are gone and version is correctly transpiled. Note that the demo misbehaves still on IE 11 (not console errors, but some misbehavior of the editor itself, but I think that's expected).

Closes: https://github.com/liferay/alloy-editor/issues/1228

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -11,6 +11,16 @@ function toAbsolute(rootRelativePath) {
 	return path.join(__dirname, rootRelativePath);
 }
 
+/**
+ * For windows compatibility, we need to use windows path separators for webpack
+ * "test" properties.
+ */
+function toTestRegExp(file) {
+	return new RegExp(
+		file.split('/').join(path.sep).replace(/\\/g, '\\')
+	);
+}
+
 const base = {
 	/**
 	 * https://webpack.js.org/configuration/entry-context/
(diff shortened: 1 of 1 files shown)


blzaugg:
Can we use `path.normalize(p)`?
https://millermedeiros.github.io/mdoc/examples/node_api/doc/path.html#path.normalize

wincent:
Ah yes, `path.normalize()` is probably better than `split()` + `join()`, but we'll still need to use `replace()` to prepare the regular expression (because we have to pass a regular expression, and not a string).

wincent:
CI failures are spurious and will be fixed by https://github.com/liferay/alloy-editor/pull/1238

julien:
Just started reviewing :)

Issue #1237, opened 2019-04-24
Don't run Travis builds with Node v12


wincent:
Travis builds blow up because node-sass is not compatible with the just-released Node v12.

wincent:
For reference, the upstream node-sass issue is: https://github.com/sass/node-sass/issues/2632

PR #1239, opened 2019-04-24
chore: make Travis use Node v11 (1.x) (#1237)

wincent:
We needed to force the use of v11 on the master branch (see #1238) because v12 was causing CI builds to blow up. This is not a problem on the 1.x branch because we were still using v8. But we should use the same version on both branches for consistency.

Related: https://github.com/liferay/alloy-editor/issues/1237


wincent:
Ugh... dropping this as not worth the hassle. Old version of gulp with old version of gulp-sass, which means old version of node-sass. Tried to use "natives@1.1.3" (and "natives@1.1.6") to solve build issues, but build is horribly broken and would probably require a bunch of other updates to make it work. Will address the duplicate lock files in a separate pull.

wincent:
Superseded by: https://github.com/liferay/alloy-editor/pull/1242

Issue #1243, opened 2019-04-24
Document IE support


wincent:
As suggested [here](https://github.com/liferay/alloy-editor/pull/1230#issuecomment-485929990), we should document the extent to which IE is supported in AlloyEditor (namely, IE11 is supported, but you may need to add "svg4everybody" and "react-app-polyfill" or similar).

wincent:
Closed by #1244.

Issue #1161, opened 2019-02-26
Content updates for alloyeditor.com


wincent:
Splitting this one off from #1091:

- [ ] Document how to use different builds; with/without CKEditor, with/without React etc (requested [here](https://github.com/liferay/alloy-editor/issues/487#issuecomment-467093606))

wincent:
@julien: I created this one so we can build up a list of stuff that needs to be in place so that we can make the most of @mwilliams2014 involvement over the next week or two.

wincent:
cc @jbalsas — if there's anything you'd like to see included in the content update please add it above.

PR #1244, opened 2019-04-24
feat: improve documentation on level of IE support (#1243)

wincent:
- Drops misleading reference to "IE9+"; we officially support "IE11" as described in #1161.
- Get rid of odious "article1", "article2" etc ids because I need to link to a specific section in "how_to_use_it.md"; we should do this everywhere but it is beyond the scope of this commit.
- Describe needed polyfills on "how_to_use_it.md".
- Provide a link to the polyfills section on "how_to_use_it.md" from "use.md".
- Add gatsby-plugin-catch-links plugin so that the aforementioned link doesn't force an unnecessary full page-reload.

Test plan: `yarn start` and check out the edited pages. Also, use the network pane to confirm that the full page-reload isn't happening.

AlloyEditor___Home
AlloyEditor___How_to_use_it_
AlloyEditor___Basic_use

Closes: https://github.com/liferay/alloy-editor/issues/1243

### DIFF
--- a/content/docs/use/how_to_use_it.md
+++ b/content/docs/use/how_to_use_it.md
@@ -6,7 +6,7 @@ order: 2
 
 ###### The section below assumes that you are using the default UI built on React. If you're using your own custom UI, these steps will be different.
 
-<article id="article1">
+<article id="add-css">
 
 ## Add AlloyEditor's CSS to the page
 
@@ -16,7 +16,7 @@ order: 2
 
 </article>
 
-<article id="article2">
+<article id="add-js">
 
 ## Add AlloyEditor's JS to the page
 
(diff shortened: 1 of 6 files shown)


blzaugg:
@wincent,

> ## Polyfilling older browsers
> 
> To work properly on older browsers such as IE11 you will need to ensure that you have the appropriate polyfills in your environment. These are:
> 
> - In order to correctly display icons, an SVG polyfill such as [svg4everybody](https://www.npmjs.com/package/svg4everybody).
> - To provide `Symbol`, which is needed by React, a polyfill like [react-app-polyfill](https://www.npmjs.com/package/react-app-polyfill).

I'm curious, from what I can tell, `svg4everybody` polyfills at the browser level; so I could just link & call `svg4everybody` in my HTML, for pages that use Alloy Editor.

But `react-app-polyfill` polyfills at the build level; I'd need to build a custom release of Alloy Editor.

Is that correct?

/cc @gregory-bretall

wincent:
No need to rebuild Alloy Editor, @blzaugg. You just need to put react-app-polyfill on the page before using it. For example, see how [the website uses react-app-polyfill](https://github.com/liferay/alloy-editor/blob/1b63f3724d3bdd821f18eff06f8bf4b430918b69/gatsby-browser.js#L2) with [a vanilla copy of Alloy Editor](https://github.com/liferay/alloy-editor/blob/1b63f3724d3bdd821f18eff06f8bf4b430918b69/src/html.js#L57).

Exactly how you get react-app-polyfill on your site is an implementation detail that will vary according to how you put your site together. For example, our website is built by Gatsby, so we just put the `require()` in the "gatsby-browser.js" file (ie. the JS to be loaded in the browser by Gatsby). It would be nice if somebody shipped a pre-built version of react-app-polyfill that you could just pull off a CDN with a script tag, but I don't know if anybody has done that.

PR #1249, opened 2019-04-26
LPS-94777 Remove format only where styleFn does not exist

sandrodw3:


julien:
Just started reviewing :)

Issue #1251, opened 2019-04-26
Still experiencing Build issues on Windows in regards to transpiling JS


gregory-bretall:
**Do you want to request a *feature* or report a *bug*?**
Report a bug

**What is the current behavior?**
When building on windows, there are still functions not getting transpiled causing problems on IE11

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
Build Alloyeditor on Windows and start it up in IE11.

**What is the expected behavior?**
AlloyEditor starts with no errors

**Which versions of alloy-editor, and which browser / OS are affected by this issue? Did this work in previous versions?**
https://github.com/liferay/alloy-editor/pull/1236 was supposed to have fixed it, but does not seem to have on my end.

wincent:
Ah, sorry about that @gregory-bretall. I had tested this on Windows and (literally) running back and forth between a Windows laptop and my own machine I broke the fix.

wincent:
Fix is included in [the v2.0.2 release](https://github.com/liferay/alloy-editor/releases/tag/v2.0.2). Thanks for reporting and fixing.

PR #1252, opened 2019-04-26
fix: backslash escaping on Windows builds (#1251)

gregory-bretall:
…ackslashes into the proper regular expression format that they need to be in

Sent in to fix https://github.com/liferay/alloy-editor/issues/1251 in relation to https://github.com/liferay/alloy-editor/pull/1236


wincent:
Thanks @gregory-bretall. I think I had it that way initially (because I tested it that way) and I screwed up in one of my edits.

Issue #1255, opened 2019-04-29
Drag and Drop Handle is Difficult to See Against Text in IE11


gregory-bretall:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
The current behavior is that the drag and drop handle in IE11 can be difficult to see against text in IE11 due to it being having low opacity and also being positioned directly above the image, allowing it to overlap on text

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
1. Start up an AlloyEditor instance in IE11
2. Add a multi-line paragraph of text
3. Add an image into the middle of the paragraph

**What is the expected behavior?**
The drag and drop handle is clearly visible again the text

**Which versions of alloy-editor, and which browser / OS are affected by this issue? Did this work in previous versions?**
It only affects IE11 as it's specific to the different drag resizer we use in IE11.

julien:
Hi @gregory-bretall,

Thanks for reporting this issue.

I tried reproducing the issue but wasn't able to see what was wrong, and although there are some changes in pr #1256 I don't see any difference in alloy-editor's behaviour compared to master.

Could you possibly send a .gif or a screenshot of this issue to help me understand.

Thanks!

gregory-bretall:
@julien I have gathered 2 sets of before and after screenshots displaying the difference in visibility before and after my change for both default AlloyEditor and AlloyEditor in Liferay (the use in Liferay is why I chose white instead of the default gray but I can change back) 
[before_and_after_drag_handles_ie11.zip](https://github.com/liferay/alloy-editor/files/3133320/before_and_after_drag_handles_ie11.zip)

PR #1256, opened 2019-04-29
fix: improve drag handler visibility on IE (#1255)

gregory-bretall:
Make the drag handler opaque, change the color to white to make it stand out more and move the handle so it is more visibly separate from the image

Fixes https://github.com/liferay/alloy-editor/issues/1255

I decided to add the styles inline as CKEditor by default already adds the styles in-line and this seemed like a cleaner implementation within the CKEditor framework than trying to implement it using real CSS.

### DIFF
--- a/src/plugins/dragresize_ie11.js
+++ b/src/plugins/dragresize_ie11.js
@@ -488,6 +488,10 @@
 				)
 					setupResizer(this);
 
+				const dragHandlerStyle = this.dragHandlerContainer.$.style;
+				dragHandlerStyle.setAttribute('backgroundColor', 'rgba(255, 255, 255, 1');
+				dragHandlerStyle.setAttribute('opacity', '1');
+
 				this.shiftState = helpers.stateShifter(this.editor);
 
 				// Add widget editing option to its context menu.


julien:
Just started reviewing :)

src/plugins/dragresize_ie11.js
```diff
@@ -485,9 +485,16 @@
 				if (
 					editor.filter.checkFeature(this.features.dimension) &&
 					editor.config.ae_dragresize_ie11_disableResizer !== true
-				)
+				) {
 					setupResizer(this);
 
+					var dragHandlerStyle = this.dragHandlerContainer.$.style;
```
wincent: (nit) Can we please make this a `const`?

src/plugins/dragresize_ie11.js
```diff
@@ -485,9 +485,16 @@
...
+					dragHandlerStyle.setAttribute('backgroundColor', 'rgba(255, 255, 255, 1');
+					dragHandlerStyle.setAttribute('opacity', '1');
+					dragHandlerStyle.setAttribute('margin-top', '-5px');
+					dragHandlerStyle.setAttribute('margin-left', '-20px');
```
wincent: Wondering if this would belong better inside `setupResizer()`? I don't know if you consider it to be related to that, but it looks that way to me.

gregory-bretall: I think it's technically separate so maybe I should move it out of the if block since I'm not sure that the resizing and the drag and drop handle are explicitly linked functionality

src/plugins/dragresize_ie11.js
gregory-bretall: Fixed

gregory-bretall:
I made some changes to the branch including moving it outside the if check being used for setupresizer (since I think the drag and drop functionality and resize functionality are supposed to be mutually exclusive) and changed it from var to const

gregory-bretall:
Note: I made another change removing positioning changes as it can potentially conflict with a possible fix to https://issues.liferay.com/browse/LPS-89596 which was originally added to liferay-ckeditor but seems to have not been transferred over to CKEditor 4.11.3. I will look into fixing that again later, but for now not touching the drag handle positioning seems like the best plan.

wincent:
> originally added to liferay-ckeditor but seems to have not been transferred over to CKEditor 4.11.3. I will look into fixing that again later, but for now not touching the drag handle positioning seems like the best plan.

Good catch. When we did the big CKEditor update we changed the build process quite dramatically, so this will be the first change that we need to integrate using [the new workflow](https://github.com/liferay/liferay-ckeditor/issues/16). I don't know what the original fix was (because the LPS just updates the version number), but I'm wondering: was it the kind of fix that could potentially be upstreamed into CKEditor itself?

In any case, in the meantime this PR looks good and low-risk so let's go ahead and merge it. Thanks for providing the screenshots to make clear what the before/after looks like.

blzaugg:
See tag [4.6.0.5](https://github.com/liferay/liferay-ckeditor/compare/4.6.0.5~3...4.6.0.5) for changes.

blzaugg:
See tag [4.6.0.5](https://github.com/liferay/liferay-ckeditor/compare/4.6.0.5~3...4.6.0.5) for LPS-89596 changes.

julien:
Hi @blzaugg unfortunately the  4.6.0.5 "tag" does not correspond to the commit you pointed, but to [this](https://github.com/liferay/liferay-ckeditor/commit/36d8b9ec8c9c09665cac3a727e0c3d32afbda87f).

I will try to find a way to have an easy patch workflow.

wincent:
This is the specific commit: https://github.com/liferay/liferay-ckeditor/commit/4d8ce3eca6d1e8ab3a7f1062b8981120a866759f

blzaugg:
Sorry, I was including the range from the previous tag, so all relevant commits were included.

PR #1258, opened 2019-04-30
feat: prevent publication with stale version number

wincent:
This should make things ~~fool~~greg-proof.

The salient bits are:

- We re-run the build as part of any "version" operation.
- We do a pre-publish check to make sure the dist files are up-to-date.

See the individual commit messages for details.


julien:
Just started reviewing :)

PR #1259, opened 2019-05-03
fix: improve drag handler visibility on IE (#1255)

gregory-bretall:
Make the drag handler opaque, change the color to white to make it stand out more and move the handle so it is more visibly separate from the image

Sent in relation to https://github.com/liferay/alloy-editor/issues/1255


wincent:
Thanks @gregory-bretall.

PR #1270, opened 2019-05-20
fix: update color for label

julien:
Please see https://issues.liferay.com/browse/LPS-95292 for more information.

That also means I'll cut a release.

### DIFF
--- a/src/assets/sass/components/ae-container/structure.scss
+++ b/src/assets/sass/components/ae-container/structure.scss
@@ -37,6 +37,7 @@
     }
 
     label {
+        color: #fff;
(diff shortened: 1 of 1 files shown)


src/assets/sass/components/ae-container/structure.scss
```diff
@@ -37,6 +37,7 @@
     }
 
     label {
+        color: white;
```
wincent: I think we usually write this as `#fff` elsewhere?

julien: Yes ... good catch

PR #1272, opened 2019-05-27
fix: add aria-label and title attributes for the image and video buttons

julien:


src/assets/lang/en.json
```diff
@@ -33,6 +33,7 @@
 	"rows": "Rows",
 	"success": "Success",
 	"table": "Insert Table",
+	"video": "Insert Video",
```
wincent: I think we're supposed to add the keys to `src/assets/lang/language.json`, not this file.

src/components/buttons/button-embed-video.jsx
```diff
@@ -46,10 +46,12 @@ class ButtonEmbedVideo extends React.Component {
 		} else {
 			return (
 				<button
+					arial-label={AlloyEditor.Strings.video}
```
wincent: s/arial/aria/

I see what you did there...

Screen Shot 2019-05-27 at 16 01 51

src/assets/lang/en.json
julien: Yes, you're right.

src/components/buttons/button-embed-video.jsx
julien: Nice good catch

Issue #1275, opened 2019-06-06
IE11 does not interpret/strips other browser's sizing via style attribute


SpencerWoo:
https://issues.liferay.com/browse/LPS-96076

introduced with https://github.com/liferay/alloy-editor/issues/898 https://github.com/liferay/alloy-editor/commit/18ac73d6f065ab0c68919b8e58247aee222e3f1a

**Current behavior**

IE removes styles from images.  The size utilizes the style in other browsers.  IE doesn't interpret/strips other browser's sizing.

**Expected behavior**

IE interprets other browser's sizing.

wincent:
Seems like this one can be closed now (thanks to #1276).

PR #1276, opened 2019-06-07
fix(ie): make IE preserve other browsers' image width styling (#1275)

SpencerWoo:
…e styling

https://github.com/liferay/alloy-editor/issues/1275

Instead of removing all the styles, keep the width (controls sizing) style

### DIFF
--- a/src/plugins/dragresize_ie11.js
+++ b/src/plugins/dragresize_ie11.js
@@ -986,7 +986,11 @@
 
 			const image = wrapper.$.querySelector('img');
 
-			image.removeAttribute('style');
+			const imageStyles = image.getAttribute('style');
+			const widthStyles = /(width:.+?;)/g.exec(imageStyles);
+			const widthStyle = widthStyles[0];
+
+			image.setAttribute('style', widthStyle);
 		}
 	}
 


src/plugins/dragresize_ie11.js
```diff
@@ -986,7 +986,11 @@
 
 			const image = wrapper.$.querySelector('img');
 
-			image.removeAttribute('style');
+			let imageStyles = image.getAttribute('style');
+			let widthStyles = /(width:.*?;)/g.exec(imageStyles);
+			let widthStyle = widthStyles[0];
```
wincent: Seems all of these `let` should be `const`, right?

Also, about that regex: `.*?` is a non-greedy match (good), but `*` implies that zero-width is acceptable and it isn't, is it? ie. `.+?` would be a more accurate specification of what we're looking for. Not that I think we'll get a zero-width value there in practice, just that I think in general it's best to keep regexes as "tight" as possible.

SpencerWoo: Yeah, they can all be `const`.  Wasn't aware that's the convention but that makes sense.

And right, `.+?` is definitely better here

SpencerWoo:
Didn't know if we were waiting for me to make those changes.

Let me know if there's anything else or when this can be merged so we can handle the LPS.  thanks

SpencerWoo:
@wincent Any update on reviewing this one?  thanks

wincent:
@SpencerWoo: We were looking this over but we had two days out due to office activities here in Madrid. @julien was testing it out and is going to merge it and cut a release today.

SpencerWoo:
awesome thanks guys

PR #1284, opened 2019-07-10
fix(firefox): broken table toolbar (#1283)

rolandpakai:
Plesae check my fix and release it in 1.5.25

Thanks, Roland

The fix was created based on this: https://github.com/liferay/alloy-editor/commit/8d2a3f7ef3e708aa091d54f73e77fa2bca01af11

Issue #1283, opened 2019-07-10
Table toolbar is broken in Firefox


rolandpakai:
https://issues.liferay.com/browse/LPS-97492

Current behavior
The tabe toolbar is not working on Firefox: can not click on the button's drop down list items

Expected behavior
Tooltip button's drop down list is usable

Affected version
1.5.24

Affected browser
Only Firefox

wincent:
Fixed by: https://github.com/liferay/alloy-editor/pull/1293

PR #1293, opened 2019-07-19
fix(firefox): broken table toolbar (#1283)

wincent:
@rolandpakai, I took your PR (https://github.com/liferay/alloy-editor/pull/1284) which was based on https://github.com/liferay/alloy-editor/commit/8d2a3f7ef3e708aa091d54f73e77fa2bca01af11, and rebased it on top of https://github.com/liferay/alloy-editor/pull/1292 (which fixes the CI failures). Also, I am targeting it at the "1.x" branch instead of "1.x-stable" (we only merge to the stable branches right before releasing). Now let's see what the CI says about this!

### DIFF
--- a/src/ui/react/src/components/base/widget-position.js
+++ b/src/ui/react/src/components/base/widget-position.js
@@ -210,15 +210,14 @@
                     top: endPoint[1] + 'px',
                     opacity: 1
                 });
+				
+				// 150ms to match transition-duration for .ae-toolbar-transition:
+				setTimeout(() => {
+					domElement.setStyles({
+						pointerEvents: '',
+					});
+				}, 150);
             });
-
-            if (domElement.$) {
-                domElement.$.addEventListener('transitionend', () => {
-                    domElement.setStyles({
-                        pointerEvents: '',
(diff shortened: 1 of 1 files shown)


src/ui/react/src/components/base/widget-position.js
```diff
@@ -210,15 +210,14 @@
                     top: endPoint[1] + 'px',
                     opacity: 1
                 });
+				
+				// 150ms to match transition-duration for .ae-toolbar-transition:
+				setTimeout(() => {
+					domElement.setStyles({
+						pointerEvents: '',
+					});
+				}, 150);
```
wincent: So on the master branch we use 100ms here, not 150ms. Do you think we should change it there as well @rolandpakai? Do you remember why you picked 100ms @julien?

I'll merge this as-is in any case, but we can loop back and do the update later if you think we should.

wincent:
Ok this is merged. Thanks @rolandpakai. I think we'll be cutting the actual releases on Monday. Probably not a good idea to rush them late on a Friday. 😁

Issue #1291, opened 2019-07-18
Bug: Insert Image is not working


pankajxeine:
bug

try to upload/insert image but insert image button is not responding.

steps
click on + button
click on insert Image

As per behavior, should open upload image window and allow to upload/insert image 

OS:- Window 10
browser: chrome/IE11

wincent:
Hi @pankajxeine. Thanks for the report. Should be fixed in "master" now.

PR #1296, opened 2019-07-22
fix: make demo insert-image button work again (#1291)

wincent:
In 8a0aaf75ee8c223a776f4b0f08d3953 we added a button (button-item-selector-image.jsx) with a duplicate "image" key, clobbering the existing "image" button (in button-image.jsx). Because of the way object literals work in JS, when we list the buttons in "src/components/buttons/index.js", the old image button is mapped to "image", then the new button is mapped over the top, overwriting it.

The new button works in Liferay DXP because we have the necessary "imageselector" command, but in the demo, it does nothing.

Fix is to pick a non-colliding name ("imageFromFile"). That fixes the demo, and it doesn't break DXP because DXP doesn't actually use the src/adapter/core.js to define what goes in each toolbar, instead using the system described here:

https://portal.liferay.dev/docs/7-1/tutorials/-/knowledge_base/t/modifying-an-editors-configuration

Test plan: `yarn build` and `yarn start`; check the demo.

Closes: https://github.com/liferay/alloy-editor/issues/1291

### DIFF
--- a/src/components/buttons/button-image.jsx
+++ b/src/components/buttons/button-image.jsx
@@ -18,7 +18,7 @@ class ButtonImage extends React.Component {
 	 * @property {String} key
 	 * @static
 	 */
-	static key = 'image';
+	static key = 'imageFromFile';
 
 	constructor(props) {
 		super(props);
@@ -87,7 +87,7 @@ class ButtonImage extends React.Component {
 	 * @method _onInputChange
 	 * @protected
 	 */
-	_onInputChange() {
+	_onInputChange = () => {
 		const inputEl = this.fileInput.current;
 
 		// On IE11 the function might be called with an empty array of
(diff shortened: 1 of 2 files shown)


src/components/buttons/button-image.jsx
```diff
@@ -87,7 +87,7 @@ class ButtonImage extends React.Component {
 	 * @method _onInputChange
 	 * @protected
 	 */
-	_onInputChange() {
+	_onInputChange = () => {
```
wincent: Perhaps unsurprisingly, the button that hadn't been used for months wasn't actually working, because `this` wasn't bound.

wincent:
FWIW, I verified that after this change, all 63(!) buttons have non-colliding names.

Issue #1297, opened 2019-07-29
IE11 does not interpret other browser's height sizing via style attribute


SpencerWoo:
https://issues.liferay.com/browse/LPS-98932

Current behavior

Other browsers use the style to set the image size.  IE removes styles from images without interpreting them -- IE sets it's own image sizing and overrides other browser's configuration.

Expected behavior

IE interprets other browser's sizing.

PR #1290, opened 2019-07-15
fix(ie): add a check for 'imageStyles' before accessing (#1289)

fortunatomaldonado:
Fixes #1289
https://issues.liferay.com/browse/LPS-98250

`imageStyles` needs to be checked before getting any information from it.

Let me know if there are any questions or comments about this.
Thank you.

Issue #1289, opened 2019-07-15
CK Resizer appears in IE11


fortunatomaldonado:
**Do you want to request a *feature* or report a *bug*?**
Bug
**What is the current behavior?**
The CK Resizer appears on images in IE11
**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
1. Start the Liferay Demo
2. Click on the image and resize the image

Actual Result: CKEditor handles appear and you are able to change the aspect ratio of the image.
**What is the expected behavior?**
Expected Result: The image will resize without an CKEditor handles appearing

**Which versions of alloy-editor, and which browser / OS are affected by this issue? Did this work in previous versions?**
2.0.7 and 1.5.24 affected in IE11

Issue #898, opened 2018-11-06
Image alignment Center CSS clashes when switching from IE to Chrome


fortunatomaldonado:
## Issue
Aligning an image in IE, the CSS is placed in the container. `style= "text-align: center";` Aligning the image in Chrome/other browsers, the CSS is placed in the image itself. `style="display: block; margin-left: auto; margin-right: auto;"` When aligning in IE and going to that content in Chrome, it is impossible to remove the CSS in the container and so the image will remain centered.

## Reproduce
I have not been able to reproduce this behavior without using Liferay content, but I do believe alloyEditor should be aware of this cross browser interaction and work accordingly.

Let me know if more information is needed and whether or not this should be resolved in AE or Liferay itself.
https://issues.liferay.com/browse/LPS-86954
Thank you.

duracell80:
This code is missing a semi colon. text-align: center. Please consider syntax errors as causing issues for customers.

fortunatomaldonado:
@duracell80 Thanks for catching the typo.

PR #1298, opened 2019-07-29
fix: make IE also preserve other browers' image height styling

SpencerWoo:
Fixes https://github.com/liferay/alloy-editor/issues/1297

Saving the image in IE sets/overrides the height value, so style height also needs to be preserved.

### DIFF
--- a/src/plugins/dragresize_ie11.js
+++ b/src/plugins/dragresize_ie11.js
@@ -989,10 +989,15 @@
 			const imageStyles = image.getAttribute('style');
 
 			if (imageStyles) {
+				const heightStyles = /(height:.+?;)/g.exec(imageStyles);
+				const heightStyle = heightStyles[0];
+
 				const widthStyles = /(width:.+?;)/g.exec(imageStyles);
 				const widthStyle = widthStyles[0];
 
-				image.setAttribute('style', widthStyle);
+				const styles = heightStyle + widthStyle;
+
+				image.setAttribute('style', styles);
 			}
 		}
 	}


src/plugins/dragresize_ie11.js
```diff
@@ -989,10 +989,15 @@
 			const imageStyles = image.getAttribute('style');
 
 			if (imageStyles) {
+				const heightStyles = /(height:.+?;)/g.exec(imageStyles);
+				const heightStyle = heightStyles[0];
+
 				const widthStyles = /(width:.+?;)/g.exec(imageStyles);
 				const widthStyle = widthStyles[0];
```
wincent: We looked at this in a bit more detail and saw that it was possible to trigger an error here because we don't check whether `widthStyles` is `null` before accessing `[0]` on it (ie. a bug that pre-existed before this PR, but we add another instance of the same problem here). To see that happen, just set a width but not a height style (or vice a versa) and see the error in the console. So @julien is preparing a follow-up that guards against that case.

wincent:
We found that in trying to repro the problem reported in [LPP-34608](https://issues.liferay.com/browse/LPP-34608) that we can't actually repro it if we comment out the code that does `image.setAttribute()`.

That makes me think this is actually a regression introduced in https://github.com/liferay/alloy-editor/pull/1290 — that code fixed another bug about an unwanted CK resize handle (cc @fortunatomaldonado), but it had the effect of blowing away all styles other than width. This PR is very similar in that it blows away all styles other than width and height; both of those seem kind of wrong, right?

Maybe we should be applying a new fix for https://github.com/liferay/alloy-editor/issues/1289 that preserves all the styles instead of just setting width. What do you think @fortunatomaldonado, @SpencerWoo?

wincent:
Tracing it back a bit further, I see that it might not be a regression introduced in #1290. I see we originally added the code to preserve width in https://github.com/liferay/alloy-editor/commit/900478aceec990b7ae323da9602d26ba8c551c30, and prior to that we just unconditionally removed the entire `style` attribute. And *that* code was added in https://github.com/liferay/alloy-editor/commit/9ad59720ad91bc697165dcaf5d64bf4beb9db090. So what we have here is fix-on-top-of-fix-on-top-of-fix and it is a bit hard to be sure that we don't break something else with anything we do here.

Once we get clarity on the required behavior here we can apply One Final Fix To Rule Them All™... right? 😁

duracell80:
If I can add a perspective distanced from the actual project code ... Only IE is setting dimensions in the HTML attributes. This is fine if everyone uses IE, but in the real world different people use different browsers to edit that content (and re-edit and collaborate).

The real once and for all fix would be to make IE interact with the style selectors and quit setting HTML attributes.

Is there a specific reason IE cannot interact with the dimensions sets in the style="" portion of the image tag?

SpencerWoo:
@wincent thanks for looking into this one.

Yeah I agree, my understanding is #898 stripped all styles and that led to a regression as some styles are required #1275.

#1275 and #1297 are built on #898 and are essentially overriding IE's width and height if other browsers have set the width and height (OB set SW (style width) and SH instead of W and H).  If we're reconsidering #898 then my understanding of that issue is the same as duracel80's comment above -- can IE set width and height via SW and SH?

I believe this question was discussed but I don't quite remember the rationale -- @fortunatomaldonado do you remember the context of #898?

fortunatomaldonado:
@SpencerWoo, 

I believe in #898, we were looking at how IE and other browsers would align images differently. For center alignment, IE would only work using text-align on the parent element <p> and not the image, while other browsers can modify the image directly. Stripping the styles was a way to remove the alignment for the image, but we did not take into consideration width and height styles.

duracell80:
I can't help but think that a set of standards would help with these image plugin concerns? Then align all HTML output to those standards? I can only think that if IE was without width in the style that would account for "margin: 0 auto; text-align: center;" not working in IE and then having to use text align (which causes issues with text being center aligned and users not having the ability to get their text back in left alignment). If IE had style width then margin zero auto should align the image?

duracell80:
Are these fixes being verified in DXP also with a series of edits cross browser to help test real world edit sequences?

wincent:
@duracell80: I don't know about #1307 (because I haven't looked at it yet) but our QA folks have specifically added a test case that involves round-tripping in (thanks @john-co) in light of the problems we've been discussing here.

wincent:
I'm going to cut v2.0.9 (https://github.com/liferay/alloy-editor/pull/1309) seeing as this one has been sitting in the master branch unreleased for a couple of weeks now, and we just merged #1307 which is a follow-up.

That will give us a baseline to test in DXP (including the cross-browser editing) and we'll create new LPS/GitHub tickets for any problems that we detect, which we can then tackle in a holistic way, informed by the discussion that we've had on this PR.

PR #1307, opened 2019-08-14
fix: check whether height and width style exist before using

SpencerWoo:
This is to fix the issues described https://github.com/liferay/alloy-editor/pull/1298#discussion_r310046875.  We're guarding against null values for height and width styles

### DIFF
--- a/src/plugins/dragresize_ie11.js
+++ b/src/plugins/dragresize_ie11.js
@@ -989,13 +989,13 @@
 			const imageStyles = image.getAttribute('style');
 
 			if (imageStyles) {
-				let styles = "";
+				let styles = '';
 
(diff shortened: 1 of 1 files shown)


wincent:
I know there is ongoing discussion in https://github.com/liferay/alloy-editor/pull/1298 about what would still be needed to make the behavior in IE really robust, especially with respect to round-tripping to/from other browsers. While this PR doesn't address *that*, at least it doesn't make things *worse* than they are, and it fixes a bug, so I think we should merge it. Thanks @SpencerWoo.

wincent:
I will push a commit on top of this fixing the formatting errors. I also pushed https://github.com/liferay/alloy-editor/pull/1308 that turns on formatting checks in CI.

PR #1313, opened 2019-08-21
fix: preserve height style for IE (1.x)

SpencerWoo:
Not certain if the 'backport' procedure is to do 1:1 commits but I've bundled the changes of this section of code into one commit here

### DIFF
--- a/src/plugins/dragresize_ie11.js
+++ b/src/plugins/dragresize_ie11.js
@@ -895,10 +895,19 @@
             var imageStyles = image.getAttribute('style');
 
             if (imageStyles) {
-                var widthStyles = /(width:.+?;)/g.exec(imageStyles);
-                var widthStyle = widthStyles[0];
+            	var styles = '';
 
-                image.setAttribute('style', widthStyle);
+            	var heightStyles = /(height:.+?;)/.exec(imageStyles);
+            	if (heightStyles) {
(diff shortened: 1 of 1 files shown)


jbalsas:
Hey @SpencerWoo, I assume this is for `7.0.x`? I think it should be safe and preferrable at this point to backport the changes to AlloyEditor v2 so we can concentrate there. Since we've had it in `7.1.x` for a while and ironed out the quirks, I'd say this is a realistic option.

What do you think?

jbalsas:
Just to be clear, I'm proposing we move `7.0.x` to `AlloyEditor@2` so we can stop maintaining the `1.x` branch independently :)

SpencerWoo:
Yes, this was for `7.0.x` -- moving `7.0.x` to `AlloyEditor@2` meets my use case and sounds good!

wincent:
> so we can stop maintaining the `1.x` branch independently

source

wincent:
By way of status update, @antonio-ortega has been looking at back-porting v2.x of AlloyEditor to DXP 7.0, which is why this one is sitting here in pending state for now.

wincent:
Backport has been done so we can effectively forget about all 1.x PRs from here on — as long as the fixes go into master/2.x, we're good.

Issue #1316, opened 2019-08-29
IE11 Drag Resizer Not Working on Top Line


gregory-bretall:
**Do you want to request a *feature* or report a *bug*?**
Report a bug

**What is the current behavior?**
When adding an image in IE11, the drag resizer is placed above the image, making it so that the drag handler appears outside of the editor when placed on the top line.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
Most easily reproducible in Liferay, see steps on https://issues.liferay.com/browse/LPP-34775.

**What is the expected behavior?**
Drag handler is visible and accessible from any point in the editor.

**Which versions of alloy-editor, and which browser / OS are affected by this issue? Did this work in previous versions?**
Affects IE11 only.

markocikos:
Closing, we no longer support IE.

PR #1317, opened 2019-08-29
fix: move drag handler to ensure visible even when image on top line (#1316)

gregory-bretall:
Closes: https://github.com/liferay/alloy-editor/issues/1316

Let me know if you think this is the correct approach or if we want to do something change the z-index of the drag handler.

### DIFF
--- a/src/plugins/dragresize_ie11.js
+++ b/src/plugins/dragresize_ie11.js
@@ -494,6 +494,8 @@
 					'rgba(255, 255, 255, 1'
 				);
 				dragHandlerStyle.setAttribute('opacity', '1');
+                dragHandlerStyle.setAttribute('left', '5px');
+                dragHandlerStyle.setAttribute('top', '5px');
 
 				this.shiftState = helpers.stateShifter(this.editor);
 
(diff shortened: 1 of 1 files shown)


wincent:
Taking this for a test-drive now on the Windows laptop here.

wincent:
The top/left doesn't seem to have any effect in my testing on IE 11 on Windows.

Sin título

I'm seeing `top: 0` and `left: 15px`, and the handle is exactly where it was before, partially obscured.

Any ideas what could explain that?

duracell80:
There has to be a different way of thinking here to break out of the difficulties. Why does it have to be at the top? Can't the position of this icon be on the right center of the image?

Then if the image is aligned right switch the position of the icon to the left center?

duracell80:
It's also very small ... I have users that are not very dexterous that struggle with drag and dropping DXP portlets with a mouse.

I have users  that edit content without a mouse because they are in a meeting on a laptop with a touchpad because they don't have a wireless mouse and the mouse is on their desk ... There needs to be some form of user feedback afforded to developers because that would be the only way to see usability improvements with this editor.

We also have many wireless mouses left in meeting rooms so if a user has lost their mouse they can only fall back into their trackpad.

wincent:
Given the doubts about this being the right approach, we're going to close this one.

PR #1318, opened 2019-09-04
fix: alignment removal (#1315)

### LINKED ISSUE #1315: IE11 Image wrong alignment after center alignment
**Do you want to request a *feature* or report a *bug*?**
Bug
**What is the current behavior?**
Image is not able to be inline with text after setting and then removing center alignment on an image in IE11
**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
1. Start Demo
2. Add a line of text and then add an image
3. Center align the image and then remove the alignment
4. Now drag and drop the image in the middle of the text

**What is the expected behavior?**
Expected Result: Image is inline with the text
Actual Result: The image is now left aligned

IE11
I believe this could be a start into looking at the way different browsers deals with alignment and styling from https://github.com/liferay/alloy-editor/pull/1298

fortunatomaldonado:
#1315 

When the alignment of an image is removed, the alignment is set back to either left or right alignment. This fix looks at the data alignment and whether the wrapper contains any alignment styles and sets it to none if necessary.
Thank you!

### DIFF
--- a/src/plugins/dragresize_ie11.js
+++ b/src/plugins/dragresize_ie11.js
@@ -349,6 +349,15 @@
 				)
 					this.data.align = 'none';
 
+				if (
+					(this.data.align === 'left' ||
+						this.data.align === 'right') &&
+					!this.wrapper.$.contains('float') &&
+					!this.oldData
+				) {
+					this.data.align = 'none';
+				}
+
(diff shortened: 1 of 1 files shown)


markocikos:
I'm closing this, as we no longer support IE11.

Issue #1314, opened 2019-08-29
Export the "mixins". Show how to export them in the proper order and document it.


mwilliams2014:

wincent:
Labeling this as a "feature" although we could just as easily classify it as a bug because it is something you used to be able to do straightforwardly in 1.x and it became difficult or impossible in 2.x.

wincent:
For the purposes of documentation, these are the existing wrapping patterns that you can see in 2.x:

```
ButtonActionStyle
  ButtonStateClasses
    ButtonStyle
      Component (x3)

ButtonActionStyle
  ButtonStyle
    Component

ButtonCfgProps
  ButtonKeystroke
    ButtonStateClasses
      Component

ButtonCfgProps
  WidgetDropdown
    WidgetFocusManager
      Component

ButtonCommand
  Component (x2)

ButtonCommand
  ButtonCommandActive
    ButtonStateClasses
      Component (x9)

ButtonCommand
  ButtonKeystroke
    ButtonStateClasses
      ButtonStyle
        Component (x3)

ButtonCommand
  ButtonStateClasses
    ButtonStyle
      Component (x5)

ButtonCommand
   ButtonStyle
     Component

ButtonKeystroke
  Component

ButtonProps
  ButtonStateClasses
    Component (x4)

ButtonProps
  ButtonStateClasses
    ButtonKeystroke
      Component

ButtonProps
  ButtonStateClasses
    ButtonStyle
      Component

ButtonStateClasses
  Component

WidgetArrowBox
  WidgetDropdown
    WidgetExclusive
      WidgetFocusManager
        WidgetPosition
          ToolbarButtons
            Component (x2)

WidgetExclusive
  WidgetFocusManager
    Component

WidgetFocusManager
  Component (x5)
```

Some but not all of them are order-sensitive (a terrible property for them to have, but fixing it basically requires a total rewrite). So looking at that list, we can simplify it to something like the following, but will need to dig though the code to determine what the real dependencies are (which is definitely a black mark against the architectural design):

```
ButtonActionStyle
ButtonStateClasses
ButtonStyle

ButtonCfgProps
ButtonKeystroke
ButtonStateClasses

ButtonCfgProps
WidgetDropdown
WidgetFocusManager

ButtonCommand
ButtonCommandActive
ButtonStateClasses
ButtonStyle

ButtonCommand
ButtonKeystroke (or ButtonStateClasses)
ButtonStateClasses (or ButtonKeystroke)

[...]

PR #1330, opened 2019-09-25
refactor: prepare to export "mixins"

wincent:
While doing auditing for https://github.com/liferay/alloy-editor/issues/1314 noticed some things that could be changed.

Audit notes (so far) are [in this gist](https://gist.githubusercontent.com/wincent/630e5550aa66c5c22e9155149e431138/raw/c641bec4dd807968313d19e7f271d5c07aa10e0b/notes.md).

Will continue with the audit tomorrow, and actually exporting the "mixins" (in quotes there because they are actually higher-order components, but we made them to replace the mixins that we used to have before migrating from `React.createClass` to ES6 classes); they're not really higher-order components in the conventional sense though, so I'm not calling them that either (instead of following the usual React convention of composition of components, they dynamically create subclasses).


julien:
Just started reviewing :)

wincent:
Thanks @julien!

wincent:
This is green at this point and I manually tested it. I am not sure if there will be more, so going to merge it. Next step will be to *actually* export the "mixins".

PR #1331, opened 2019-09-30
feat: make it possible to build custom buttons that use "mixins"

wincent:
Please see the commit messages for details. The first one contains the juicy bit:

> Exposes:
>
> - The 2.x HOCs as `AlloyEditor.Base.ButtonCommand` (etc) so that people can write custom buttons using the new ES6-class based style.
> - The 1.x mixins as `AlloyEditor.Compat.ButtonCommand` (etc), for people who have custom buttons written using the old (`React.createClass`) style.
>
> The obvious downside here is that we're shipping more code now, increasing the bundle size and overall complexity, but we think it's worth it because we are prepared to pay quite a high cost in the name of maintaining backwards compatibility.
>
> Having said that, we are not going to go out of our way to document the `Compat` mixins at this time, so as not to overwhelm users; this is just something we want to have up our sleeve to be able to offer it if the need arises (and it may not arise, given that custom buttons probably aren't often implemented).

Closes: https://github.com/liferay/alloy-editor/issues/1314

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,7 +14,6 @@ module.exports = {
 	},
 	extends: ['liferay/react'],
 	rules: {
-		'require-jsdoc': 'warn',
 		'new-cap': [
 			'error',
 			{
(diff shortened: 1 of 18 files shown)


wincent:
I'm going to write a brief technical document now that @mwilliams2014 can use as a basis for explaining how this is all supposed to work. Will add a link to it once it is done.

wincent:
Add docs here: https://github.com/liferay/alloy-editor/issues/1314#issuecomment-536532915

PR #1333, opened 2019-10-08
docs: remove CHANGELOG-1.x.md

wincent:
Its contents are already in CHANGELOG.md. Back when we were still using github_changelog_generator, we thought that having the 1.x and 2.x histories in separate files would make it easier to reliably update the changelogs, so we split this file out in 609795f04140cfd910.

But it looks like that around 67ddb55458357c599 we started using liferay-changelog-generator, which doesn't suffer from the issues we were facing with the old tool, and that means we can just delete this file.


wincent:
Trivial, so going to merge.

PR #1336, opened 2019-10-21
fix(button-spacing): adds missing EditorContext ref

jbalsas:
Hey @wincent, @julien, just saw this in reference to https://issues.liferay.com/browse/LPS-102718 which is actually throwing an error there... it might be a different thing though, but this at least looked off...

Also, I assume there's something wrong in the description here, as it states:

```
* The ButtonSpacing class provides functionality for changing text color in a document.
```

We might need to update that :D


wincent:
Thanks @jbalsas. I don't know if this will fix the linked bug, but the change definitely looks correct.

PR #1338, opened 2019-10-25
chore: update linting

wincent:
- Updates linting dependencies and applies autofixes.
- Maintains copyright headers with notice/notice plugin.
- Lints previously unlinted files.


julien:
Just started reviewing :)

PR #1340, opened 2019-10-28
fix: include CNAME in public directory

wincent:
If you look at [the commits on the gh-pages branch to the "CNAME" file](https://github.com/liferay/alloy-editor/commits/0bc43b350037a12132f7178c292ec90a964f40f3/CNAME
) you'll see that it was getting blown away on every site build and we've had to manually restore it each time.

This commit includes it in the "public" directory, which is used by the deploy.sh script as a basis for creating commits on the gh-pages branch, so that the file will stick around on every build.

Test plan: run `deploy.sh` (no need to actually `git push` the result) and inspect the generated commit; see that it did not delete the CNAME file from the `gh-pages` branch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,5 +59,8 @@ typings/
 .env
 
 .cache/
-public
+/public/*
 yarn-error.log
+
+# Used by gh-pages to associate site with alloyeditor.com domain.
+!/public/CNAME
--- a/public/CNAME
+++ b/public/CNAME
@@ -1,0 +1,1 @@
+alloyeditor.com


wincent:
Just to reflect the result of the test plan a bit more; here's the end of the `./deploy.sh` output:

```
Creating commit object on gh-pages branch
commit 3d19a82f6abf13b4a3c30746950470e52f9857cf (gh-pages)
Author: Greg Hurrell <greg.hurrell@liferay.com>
Date:   Mon Oct 28 09:34:01 2019 +0100

    Autogenerated public/ content from 256a9845
Restoring index to prior state

Now run 'git push upstream gh-pages'
```

And [this is the output](https://gist.github.com/wincent/b5d04b738a2ed395f0888d85254aeed6) of `git whatchanged 3d19a82f6abf13b4a3c30746950470e52f9857cf -n 1`; note how it leaves the CNAME intact, unlike what used to happen (eg. see [a78b7474f7492](https://github.com/liferay/alloy-editor/commit/a78b7474f7492dd7a5dc94a57ee0be54f8103cf5#diff-adc4bfdb0829dae99e3699393e3fbaa4)).

Seeing as this is pretty trivial, I'm going to go ahead and merge it.

PR #1341, opened 2019-10-28
feat: add some useful worktree-related .gitignore entries

wincent:
Makes it convenient to have "1.x" and "gh-pages" checkouts nested in a repo clone (eg. created with `git worktree add gh-pages gh-pages` etc).


wincent:
I found this useful while creating/testing #1340, for example.

PR #1360, opened 2020-01-10
fix: avoid adding target attribute when default is selected

antonio-ortega:
Hi guys,

Can you review it, please?

Thanks.


julien:
Re-submitted here https://github.com/liferay/alloy-editor/pull/1361
I just updated the tests as well
Thanks @antonio-ortega

PR #1362, opened 2020-01-10
chore: prepare for release v1.5.27

julien:


julien:
leaving a comment here for myself later
Travis / Saucelabs are so unstable I'm thinking of running github actions to run the tests when we push changes.

Issue #1367, opened 2020-01-28
Demo > Lack of padding around images


duracell80:
In The demo. There is none. There should be at least a default amount of padding around images until the editor gives the user the ability to select on the image toolbar the amount of padding that is appropriate for their content ... then a default wouldn't be required.

wincent:
Previously reported as https://github.com/liferay/alloy-editor/issues/844, which is a bit more general than this one.

Issue #1368, opened 2020-01-28
Image overflows bounds of content


duracell80:
Image overflows bounds of content. I think because I can easily show this issue on both the demo and in DXP itself (multiple versions) ... I think it points to a serious lack of testing.

![image-overflow-content](https://user-images.githubusercontent.com/18121278/73279441-c008ae80-41b2-11ea-80ba-0bbc2b9ad29b.jpg)
![image-overflow-paragraph-fragment](https://user-images.githubusercontent.com/18121278/73279442-c008ae80-41b2-11ea-96f3-9713df6748e8.jpg)

duracell80:
All i'm pointing out is that even in the demo the editor doesn't work right and that's translating to failures in DXP too.

PR #1373, opened 2020-02-14
fix: Lack of padding around images

georgel-pop-lr:
Fix Lack of padding around images #1367

### DIFF
--- a/src/plugins/imagealignment.js
+++ b/src/plugins/imagealignment.js
@@ -28,23 +28,53 @@ if (!CKEDITOR.plugins.get('ae_imagealignment')) {
 	];
 
 	/**
-	 * Necessary styles for the center alignment
+	 * Necessary styles for the left|center|right alignment
 	 * @type {Array.<Object>}
 	 */
-	const CENTERED_IMAGE_STYLE = [
-		{
-			name: 'display',
-			value: 'block',
-		},
-		{
-			name: 'margin-left',
-			value: 'auto',
-		},
-		{
-			name: 'margin-right',
-			value: 'auto',
-		},
-	];
+	const IMAGE_STYLE_ALIGNMENT = {
+		[IMAGE_ALIGNMENT.LEFT]: [
+			{
+				name: 'display',
+				value: 'inline-block',
+			},
+			{
+				name: 'float',
+				value: 'left',
+			},
+			{
+				name: 'margin-right',
+				value: '1.2rem',
+			},
+		],
+		[IMAGE_ALIGNMENT.CENTER]: [
+			{
+				name: 'display',
+				value: 'block',
+			},
+			{
+				name: 'margin-left',
+				value: 'auto',
+			},
+			{
+				name: 'margin-right',
+				value: 'auto',
+			},
+		],
+		[IMAGE_ALIGNMENT.RIGHT]: [
+			{
+				name: 'display',
+				value: 'inline-block',
+			},
+			{
+				name: 'float',
+				value: 'right',
+			},
+			{
+				name: 'margin-left',
+				value: '1.2rem',
+			},
+		],
+	};
 
 	/**
 	 * Retrieves the alignment value of an image.
@@ -64,7 +94,9 @@ if (!CKEDITOR.plugins.get('ae_imagealignment')) {
 		}
 
 		if (!imageAlignment) {
-			let centeredImage = CENTERED_IMAGE_STYLE.every(style => {
+			let centeredImage = IMAGE_STYLE_ALIGNMENT[
+				IMAGE_ALIGNMENT.CENTER
+			].every(style => {
 				let styleCheck = image.getStyle(style.name) === style.value;
 
 				if (!styleCheck && style.vendorPrefixes) {
@@ -83,18 +115,20 @@ if (!CKEDITOR.plugins.get('ae_imagealignment')) {
 				const imageContainer = image.$.parentNode;
 
 				if (imageContainer.style.textAlign == IMAGE_ALIGNMENT.CENTER) {
-					CENTERED_IMAGE_STYLE.forEach(style => {
-						image.setStyle(style.name, style.value);
+					IMAGE_STYLE_ALIGNMENT[IMAGE_ALIGNMENT.CENTER].forEach(
+						style => {
+							image.setStyle(style.name, style.value);
 
-						if (style.vendorPrefixes) {
-							style.vendorPrefixes.forEach(vendorPrefix => {
-								image.setStyle(
-									vendorPrefix + style.name,
-									style.value
-								);
-							});
+							if (style.vendorPrefixes) {
+								style.vendorPrefixes.forEach(vendorPrefix => {
+									image.setStyle(
+										vendorPrefix + style.name,
+										style.value
+									);
+								});
+							}
 						}
-					});
+					);
 					centeredImage = true;
 					imageContainer.style.textAlign = '';
 				}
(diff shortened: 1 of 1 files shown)


wincent:
@georgel-pop-lr: to fix the CI failure, please run the `yarn format` task and either:

- `git commit --amend -p` the changes and force push (`git push -f`) your branch; or:
- Add another commit on top (`git commit -p -m "style: fix formatting"`) and push that.

Thanks!

julien:
@georgel-pop-lr this works fine on Chrome/Firefox but not in IE 11, as you can see in this screenshot 

ie11

Could you please have a look into this when you have time? 
Thanks

georgel-pop-lr:
> @georgel-pop-lr this works fine on Chrome/Firefox but not in IE 11, as you can see in this screenshot
> 
> ie11
> 
> Could you please have a look into this when you have time?
> Thanks

@julien this is the bug related to ie and edge that I was telling you about but is related to list next to image only, after talking with @antonio-ortega we agreed to open a separate bug, but it can be also fixed in this pr. we need to add to the ol element a style 'display: inline-block;'
what I'm not sure in what file can I add this style to the list elements, maybe you can guide me.

julien:
@georgel-pop-lr I'm not sure I understand but yes let's talk about it

julien:
@gregory-bretall I talked about this with @antonio-ortega and everything is clear.
I'll merge this and wait for the IE related fix before we cut a new release.
Thanks!

jbalsas:
Apologies for the late intervention. We've never agreed to changes like this because inlining styles in content makes it nearly impossible for it to be styled afterwards and reused in different apps.

We've always argued that the editor should add the minimal necessary markup and that apps should be the ones properly styling as they saw fit.

That being said, feels like an uphill battle because we keep getting reports, so we might as well fix it and deal with the problem of content not being easily styled if it comes up... 🤷‍♂

duracell80:
This can't be true because the base CK Editor inserts images with padding.

It's a choice to make the ontent look ugly and unstyled. The editor must have base level styling otherwise there's no point to this software.

And please stop applying this policy to everything like quotes. What are we going to do next remove the styling from bullet points?

This is a choice that is bad choice for the editor.

duracell80:
Markup and styling are different things.

You can't have images and text with no spacing  gutter between them it's not acceptable.

jbalsas:
Hey @duracell80 !

> Formatting is one of the biggest complaints I hear about this editor. Please listen.

We are, that's why this is being merged and also we're working other angles to fix this long term.

> And please stop applying this policy to everything like quotes. Blockquotes aren't styled either. What are we going to do next remove the styling from bullet points?

I feel you're the one that does not listen to me. I've told you many times you're conflating different issues here. Blockquotes have nothing to do with this editor. Some apps decide to strip the tags in their config (blogs) and bootstrap resets the `blockquote` tag making it look unstyled, not the editor. We already brought back default styling even though it being removed had nothing to do with AlloyEditor.

> This is a choice that is bad choice for the editor. This editor is not allowing content editors to create content that looks nice.

I agree with this. We need to make it easier. That doesn't mean we have to stop doing the other thing.

> I've noticed lack of padding on CK Editor in Liferay too so there's definitely an orchestrated effort to remove styling from content. It's just wrong, it can't be right to go down this path of removing styling.

There's no such thing, please don't feed conspiracy theories, because there's no such thing. This is how CKEditor has always worked and we just inherited from it. If something has changed along the way, we might've missed it. Could you point us to where in the OOTB CK4 demos you see the editor adding paddings?

For example, if you check the [Full Preset Demo](https://ckeditor.com/docs/ckeditor4/latest/examples/fullpreset.html) you'll see that padding is added via an additional [contents.css](https://cdn.ckeditor.com/4.13.1/full/contents.css?t=JB9C) and is not part of the default editor offering.

> Your styling issues can be overcome by not using style= and using class= instead. You're not helping apps style their content independently by not using class.
> 
> Aka class="padding-15 align-left size-medium border-thin".

duracell80:
Images are the biggest PITA in Alloy Editor, period.

Please do something to address the lack of styling in the editor. I really badly want to go off topic because these issues are self imposed (and yes in the editor a blockquote should look different to the text around it) . But staying on topic ...

Yes CK Editor by default has image padding and resize handles and the ability to display a caption and the ability to resize an image by typing in numbers.

padding-in-ck.

Yet take a look at calendar in 7.2 which is using CK Editor (I would think a copy or fork that Liferay have messed with) ... and I'm not spinning conspiracy, how can it be conspiracy if you've said the aim is to remove styling and strip it all back to barebones so we can focus on our content?

No padding and no resize handles either?
padding-images-ckeditor

The truth is that there's no editor configuration at all in Liferay that even remotely equates to an editing experience that works. Even electing to use CK Editor in Liferay renders you with less features and less styling and just less. I don't understand why this self imposed problem has been allowed to continue for so long.

You've said the aim is to limit markup, I agree, but that shouldn't mean remove all styling. It can;t, that's just the highest level of insanity. The better approach would be to add classes to the HTML. Think about it this way, the HTML being produced in people's content stays there for years and I as a style guide writer am powerless to style their content because of inline styles, constant HTML that is contradicting itself on alignment and image sizing.

You've taken too much away.

duracell80:
I totally agree with you "inlining styles in content makes it nearly impossible for it to be styled afterwards and reused in different apps."

What I cannot accept is that this should translate to removing all styles, because styling should be added using classes. You should not be aligning images with inline style, you should be adding "align-left" as a class.

jbalsas:
Hey @duracell80, looking at https://ckeditor.com/ckeditor-4/demo/, which I assume is where you took your screenshot:

Screen Shot 2020-02-18 at 18 20 59

Margin is coming from `figure` which (other than the standard browser style) is overridden by `body.article-editor figure` in [contents.css](https://ckeditor.com/assets/libs/ckeditor4/4.13.1/contents.css?t=JBAB) which again, is not part of CKEditor and `figure` has no additional classes whatsoever that get used.

If you add your own image and set `alignment` to `left`, this is the result:

Screen Shot 2020-02-18 at 18 25 23

Only when you caption it, because it gets wrapped in a `figure` tag the image gets some margin to breathe.

jbalsas:
> You're actually making our jobs much harder, because we somehow have to figure out how to add our own style sheets to your editor?? (I'm not talking about content view, I can see that it's reasonable for me to style the content) ... but you're refusing to even put basic styles around blockquotes in the editor view?

We're not refusing. We've added them.

duracell80:
It seems to be splitting hairs at this point. There is a lack of padding (actually margin) around an image that makes the text run to close to the image and it leads some users to use tables instead.

Is there an acceptance that some basic styling needs to be provided?

jbalsas:
> It seems to be splitting hairs at this point. There is a lack of padding (actually margin) around an image that makes the text run to close to the image and it leads some users to use tables instead.
> 
> Is there an acceptance that some basic styling needs to be provided?

**Not at all, we're in agreement here**. I 100% support your claim and think it's paramount that we address this and provide the right content creation experience.

What I don't think can be left unanswered are **inaccurate claims on how CKEditor works** that makes it look like we're intentionally making things worse. Same way for **issues we've agreed to fix (and fixed)** and are not a caused by AlloyEditor but by something else yet you keep bringing up on unrelated issues.

Again, 100% in agreement that easily write good-looking content needs to be our priority. And as I explained and I think you nailed it in the past, it's a problem that extends far beyond the editor. We're working to fix this holistically, because there's many actors involved

duracell80:
I think the root issue is that there are blinkers on horses. I think all you guys are able to see is that by providing styling you're making it hard to style the content.

Draw a line. There are two views. I as a theme de eloper should be responsible as for the styling of the content.

You guys should provide styling in the editor view and keep it out of the markup (but that doesn't mean provide zero styling).

And again blockquote is a brilliant example of this failure. I can put in theme styling to make a really nice block quote yet all the editor can do is indent the text slightly and that's all. It's pretty sad.

Particularly because CKEditor by default contains much more styling in the editor. This is not a false claim.

duracell80:
Pragmatically what can we do?

Simple. You guys provide classes on the markup for even aligning an image. Then provide a text area in the control panel of DXP to upload an editor stylesheet.

If you guys won't style the content in the editor (not the content on the page) give us a way to fix your product for ourselves so we can inject padding into an image in the editor and then you guys don't have to deal with such work.

duracell80:
There are no inaccurate claims. When you go to the CKEditor demo there is padding around the image.

Regardless of if that is a figure or an image the point is in the demo that loads up the image is not up against the text and that's what I think styled content looks like.

duracell80:
There is a serious issue with images in Alloy Editor that isn't related to implementation in DXP but then show up in DXP.

Like ... https://github.com/liferay/alloy-editor/issues/1368

If the editor demo can't be for for purpose how can we have any hope for DXP?

PR #1377, opened 2020-02-17
fix: show image alt text in alt text input field in IE

mateomustapic:


julien:
@mateomustapic thanks for looking at this, tested on IE11, Firefox and Chrome and it works fine

src/components/buttons/button-accessibility-image-alt.jsx
```diff
@@ -29,7 +29,11 @@ class ButtonAccessibilityImageAlt extends React.Component {
 			.getSelection()
 			.getSelectedElement();
 
-		const imageAlt = (element && element.getAttribute('alt')) || '';
+		const imageElement = element.findOne('img');
+
+		const imageAlt = imageElement
+			? imageElement && imageElement.getAttribute('alt')
```
wincent: You don't need the `imageElement &&` here; you just checked `imageElement` on the line above.

I sent a fix for this here: https://github.com/liferay/alloy-editor/pull/1379

PR #1382, opened 2020-02-24
chore: update for compliance with current Outbound Licensing Policy

wincent:
This change is analogous to the one I made in:

- https://github.com/liferay/eslint-config-liferay/pull/151
- https://github.com/liferay/liferay-npm-tools/pull/394
- https://github.com/liferay/liferay-js-themes-toolkit/pull/441
- https://github.com/liferay/clay/pull/2947

It updates the license header format to match the latest policy:

https://grow.liferay.com/excellence/Liferay+Outbound+Licensing+Policy

Did this by updating `copyright.js` and running `yarn lint:fix`. I used a Vim macro to strip out the original headers. I then edited the template to use a dynamic `<%= YEAR %>` instead of hard-coded 2014, so that any subsequently created files will use the year of creation.

Adds a copy of the license to `LICENSES/LGPL-3.0-or-later.txt` as required by the policy too.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
+/copyright.js
 src/__generated__/*
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,5 @@
 /**
- * © 2014 Liferay, Inc. <https://liferay.com>
- *
+ * SPDX-FileCopyrightText: © 2014 Liferay, Inc. <https://liferay.com>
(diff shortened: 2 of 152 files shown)


.eslintignore
```diff
@@ -1 +1,2 @@
+/copyright.js
```
wincent: Need to ignore the template itself so that it doesn't complain about finding `YEAR` when it is looking for a number like 2020, 2021 etc.

wincent:
Just comment changes, so merging.

PR #1383, opened 2020-02-24
fix: make REGEX_URL validate links and stop validating non links

julien:
see https://issues.liferay.com/browse/LPS-99704

### DIFF
--- a/src/plugins/autolink.js
+++ b/src/plugins/autolink.js
@@ -24,8 +24,7 @@ if (!CKEDITOR.plugins.get('ae_autolink')) {
 
 	const REGEX_LAST_WORD = /[^\s]+/gim;
 
-	// Seen at
-	// https://stackoverflow.com/questions/5717093/check-if-a-javascript-string-is-a-url/49849482#5717133/
+	// Seen at https://stackoverflow.com/a/5717133/2103996
 	const REGEX_URL =
 		'^(https?:\\/\\/)?' + // protocol
 		'((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|' + // domain name
--- a/src/plugins/autolink.js
+++ b/src/plugins/autolink.js
@@ -25,7 +25,12 @@ if (!CKEDITOR.plugins.get('ae_autolink')) {
 	const REGEX_LAST_WORD = /[^\s]+/gim;
 
 	const REGEX_URL =
-		'((([A - Za - z]{ 3, 9}: (?: \\/\\/)?)(?:[-;:&=\\+\\$,\\w]+@)?[A-Za-z0-9.-]+|(https?\\:\\/\\/|www.|[-;:&=.\\+\\$,\\w]+@)[A-Za-z0-9.-]+)((?:\\/[\\+~%\\/.\\w-_]*)?\\??(?:[-\\+=&;%@.\\w_]*)#?(?:[\\w]*))((.*):(\\d*)\\/?(.*))?)';
+		'^(https?:\\/\\/)?' + // protocol
+		'((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|' + // domain name
+		'((\\d{1,3}\\.){3}\\d{1,3}))' + // OR ip (v4) address
+		'(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*' + // port and path
+		'(\\?[;&a-z\\d%_.~+=-]*)?' + // query string
+		'(\\#[-a-z\\d_]*)?$';
 
 	const REGEX_EMAIL = /[a-zA-Z0-9._-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,4}/i;
 
--- a/src/plugins/autolink.js
+++ b/src/plugins/autolink.js
@@ -24,6 +24,8 @@ if (!CKEDITOR.plugins.get('ae_autolink')) {
 
 	const REGEX_LAST_WORD = /[^\s]+/gim;
 
+	// Seen at
+	// https://stackoverflow.com/questions/5717093/check-if-a-javascript-string-is-a-url/49849482#5717133/
 	const REGEX_URL =
 		'^(https?:\\/\\/)?' + // protocol
 		'((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|' + // domain name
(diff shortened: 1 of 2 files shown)


test/plugins/test/autolink.js
```diff
@@ -109,13 +109,15 @@ describe('AutoLink', function() {
 
 	it('should not create a link if last word is not a valid URL', function() {
 		testLink.call(this, {
-			expected: '<p>invalid link ww.liferay.com</p>',
+			expected:
+				'<p>invalid link <a href="http://ww.liferay.com">ww.liferay.com</a></p>',
```
wincent: So none of these tests make sense any more, do they? They say "should not create a link" but evidently we now do because `ww.liferay.com` *does* look like a URL.

src/plugins/autolink.js
```diff
@@ -25,7 +25,12 @@ if (!CKEDITOR.plugins.get('ae_autolink')) {
 	const REGEX_LAST_WORD = /[^\s]+/gim;
 
 	const REGEX_URL =
-		'((([A - Za - z]{ 3, 9}: (?: \\/\\/)?)(?:[-;:&=\\+\\$,\\w]+@)?[A-Za-z0-9.-]+|(https?\\:\\/\\/|www.|[-;:&=.\\+\\$,\\w]+@)[A-Za-z0-9.-]+)((?:\\/[\\+~%\\/.\\w-_]*)?\\??(?:[-\\+=&;%@.\\w_]*)#?(?:[\\w]*))((.*):(\\d*)\\/?(.*))?)';
+		'^(https?:\\/\\/)?' + // protocol
+		'((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|' + // domain name
+		'((\\d{1,3}\\.){3}\\d{1,3}))' + // OR ip (v4) address
+		'(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*' + // port and path
+		'(\\?[;&a-z\\d%_.~+=-]*)?' + // query string
+		'(\\#[-a-z\\d_]*)?$';
```
wincent: Splitting this up over lines and adding comments is a great idea.

If you got this from somewhere it would be good to include a link to the source.

julien: :+1:

test/plugins/test/autolink.js
julien: I modified the test descriptions, tell me what you think, it still might seem strange in some cases  :man_shrugging:

src/plugins/autolink.js
```diff
@@ -24,6 +24,8 @@ if (!CKEDITOR.plugins.get('ae_autolink')) {
 
 	const REGEX_LAST_WORD = /[^\s]+/gim;
 
+	// Seen at
+	// https://stackoverflow.com/questions/5717093/check-if-a-javascript-string-is-a-url/49849482#5717133/
```
wincent: Pro-Tip™: The "share" link on Stack Overflow answers will give you a shortlink:

https://stackoverflow.com/a/5717133/2103996

test/plugins/test/autolink.js
```diff
@@ -107,18 +107,17 @@ describe('AutoLink', function() {
 		assert.equal(data, '<p><span>@test</span> </p>');
 	});
 
-	it('should not create a link if last word is not a valid URL', function() {
+	it('should to try to create a link from the last word', function() {
```
wincent: Grammar seems a bit off here ("should to try to"), and I think our linter is going to complain about [starting an `it` with should](https://github.com/liferay/eslint-config-liferay/blob/master/plugins/eslint-plugin-liferay/docs/rules/no-it-should.md).

How about just:

```js
it('creates a link when the last word is a URL', function() {
```

src/plugins/autolink.js
julien: Prop-Tip accepted thanks!

test/plugins/test/autolink.js
```diff
@@ -107,18 +107,17 @@ describe('AutoLink', function() {
...
 		testLink.call(this, {
-			expected:
-				'<p>invalid link <a href="http://ww.liferay.com">ww.liferay.com</a></p>',
-			html: 'invalid link ww.liferay.com { }',
+			expected: '<p>invalid link http:ww.liferay.com</p>',
+			html: 'invalid link http:ww.liferay.com { }',
```
wincent: This one actually is invalid, so I think we need to keep it in a separate `it` block (ie. the old one that was there before):

```js
it('should not create a link if last word is not a valid URL', function() {
```

test/plugins/test/autolink.js
```diff
@@ -130,8 +129,14 @@ describe('AutoLink', function() {
...
-				'<p>invalid link <a href="http://ww.liferay.com">ww.liferay.com</a></p><p>text</p>',
-			html: '<p>invalid link ww.liferay.com</p>{ }<p>text</p>',
+				'<p>link <a href="http://ww.liferay.com">ww.liferay.com</a></p><p>text</p>',
+			html: '<p>link ww.liferay.com</p>{ }<p>text</p>',
+			keyCode: KEY_ENTER,
+		});
+
+		testLink.call(this, {
+			expected: '<p>invalid link eeeewwww</p><p>text</p>',
+			html: '<p>invalid link eeeewwww</p>{ }<p>text</p>',
```
wincent: And this one would go in the same `it` along with the other actually-invalid one.

test/plugins/test/autolink.js
```diff
@@ -205,15 +210,33 @@ describe('AutoLink', function() {
...
+
+		testLink.call(this, {
+			expected: '<p>invalid link wwwwwwwww</p>',
+			html: '<p>invalid link wwwwwwwww { }</p>',
+			keyCode: KEY_SPACE,
+		});
+
+		testLink.call(this, {
+			expected: '<p>invalid link www www www www</p>',
+			html: '<p>invalid link www www www www { }</p>',
```
wincent: Also would move inside "invalid" `it`.

test/plugins/test/autolink.js
julien: Much better thanks

test/plugins/test/autolink.js
julien: Yes, I'm gonna go ahead a put valid/invalid link tests in their respective "it" blocks

src/plugins/autolink.js
wincent: `React.Prop-Tip`

PR #1387, opened 2020-03-12
fix: check that the selected element is valid

julien:
Sometimes when interacting images in the editor,
the "selected element" can be null, and if this is the
case this code will fail.
An interesting this to note is that this code runs
although we haven't interacted with this
'button-accessibility-image-alt.jsx' component (probably
due to the fact these components are all HOCs)

### DIFF
--- a/src/components/buttons/button-accessibility-image-alt.jsx
+++ b/src/components/buttons/button-accessibility-image-alt.jsx
@@ -20,15 +20,17 @@ class ButtonAccessibilityImageAlt extends React.Component {
 
 	static key = 'imageAlt';
 
-	componentDidMount() {
-		const editor = this.props.context.editor;
-		const nativeEditor = editor.get('nativeEditor');
-		const selection = nativeEditor.getSelection();
+	constructor(props) {
+		super(props);
 
-		const element = selection.getSelectedElement();
-		const startElement = selection.getStartElement();
+		const selection = props.context.editor
+			.get('nativeEditor')
+			.getSelection();
 
-		this._element = element || startElement;
+		const element =
+			selection.getSelectedElement() || selection.getStartElement();
+
+		this._element = element;
 
 		const imageElement = this._element.findOne('img');
 
--- a/src/components/buttons/button-accessibility-image-alt.jsx
+++ b/src/components/buttons/button-accessibility-image-alt.jsx
@@ -23,10 +23,17 @@ class ButtonAccessibilityImageAlt extends React.Component {
 	constructor(props) {
 		super(props);
 
-		const element = props.context.editor
-			.get('nativeEditor')
-			.getSelection()
-			.getSelectedElement();
+		const editor = props.context.editor;
+
+		const nativeEditor = editor.get('nativeEditor');
+
+		const selection = nativeEditor.getSelection();
+
+		const element = selection.getSelectedElement();
+
+		if (!element) {
+			return;
+		}
 
 		const imageElement = element.findOne('img');
 
--- a/src/components/buttons/button-accessibility-image-alt.jsx
+++ b/src/components/buttons/button-accessibility-image-alt.jsx
@@ -36,7 +36,7 @@ class ButtonAccessibilityImageAlt extends React.Component {
 
 		const imageAlt = imageElement
 			? imageElement.getAttribute('alt')
-			: this._element && this._element.getAttribute('alt');
+			: this._element.getAttribute('alt');
 
 		this.state = {
 			imageAlt,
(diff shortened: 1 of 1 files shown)


wincent:
> probably due to the fact these components are all HOCs

So I don't think the HOCs have anything to do with it. The code is running in the `constructor` (ie. before the button mounts), and at that time it seems quite possible that there is no selection. Now, I don't know *why* we are constructing a button when there is no selection — as far as I know, this button [should only be constructed if there is an image selection](https://github.com/liferay/liferay-portal/blob/73711d2c28e400b5bb96cba031a8d27372555507/modules/apps/frontend-editor/frontend-editor-alloyeditor-web/src/main/java/com/liferay/frontend/editor/alloyeditor/web/internal/editor/configuration/AlloyEditorConfigContributor.java#L210-L223).

When you compare this button to the others in that list, it's the only one running code in the constructor, and looking at the Git history, it's been that way since [it was copied here from liferay-portal](https://github.com/liferay/alloy-editor/commit/d1b061729f9ca4f9476059c6f9b9bb53dc8cc32e). (It is also the only one which isn't wrapped in any of the HOCs, but I don't think that has anything to do with it in this case.)

So here's an idea, could you see what happens when you refactor this to not do this work in the constructor? (ie. just do it lazily). That might not work, but it would be interesting to find out. The other thing I notice is that none of the other buttons in this group (imageLeft, imageRight, imageCenter, linkBrowse), none of them are calling `getSelection()`... 🤔Maybe there is something edge-casey here that could lead there to be a selection but no selected element. Maybe a non-`Element` `Node`? If you debug and log the selection, what do you see? The evidence there might provide you with a way to be more confident about the fix.

In any case, some questions/observations:

- What state is the button in as a result of this early `return`? You won't have assigned `this.state`, so it will be empty — you'll still render, but `this.imageAlt` will be `undefined`.
- Looking at the code here, we store `element` in state but never change it, so it shouldn't be in state (it could be assigned to `this._element` if you wanted to cache it, though; for example, in `componentDidMount`).

julien:
> as far as I know, this button should only be constructed if there is an image selection.

I don't know about "constructed", but this button is visible when there is an image selection - and when clicking on the button, the previous image selection is lost.

Maybe that when clicking the link button we can detect if there is an exising image selection and "[lock](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_dom_selection.html#method-lock)" it to make sure it doesn't get lost - which is what I'm going to try.

> In any case, some questions/observations:

> What state is the button in as a result of this early return? You won't have assigned this.state, so it will be empty — you'll still render, but this.imageAlt will be undefined.

Yes, I'll make sure `imageAlt` is `undefined` when `render` is called 
 
> Looking at the code here, we store element in state but never change it, so it shouldn't be in state (it could be assigned to this._element if you wanted to cache it, though; for example, in componentDidMount).

Yes that also makes sense.

julien:
@wincent , I checked the CI errors, and I'm actually surprised it's failing because of the lint errors (at least from what I saw) if I remember correctly, these lint errors haven't been fixed in all files or am I missing something?

src/components/buttons/button-accessibility-image-alt.jsx
```diff
@@ -20,29 +20,23 @@ class ButtonAccessibilityImageAlt extends React.Component {
 
 	static key = 'imageAlt';
 
-	constructor(props) {
-		super(props);
-
-		const editor = props.context.editor;
-
+	componentDidMount() {
+		const editor = this.props.context.editor;
```
wincent: As per our Slack chat, if this is not a timing issue, and moving the code from the `constructor` to `componentDidMount` doesn't change the behavior, then there is no point in moving it. Is there any reason why you couldn't make this same change (calling `getStartElement()`)  in-place where the code used to be?

As it is currently written, you are reassigning to `this.state` in `componentDidMount()`, which is wrong. The only legit place to assign like that is in the `constructor`. Elsewhere it has to be via `this.setState`, and even then it is not allowed in all places:

1799171_10201997263550081_674283271_o

wincent:
> @wincent , I checked the CI errors, and I'm actually surprised it's failing because of the lint errors (at least from what I saw) if I remember correctly, these lint errors haven't been fixed in all files or am I missing something?

This one is legit:

```
39:3   error    Do not mutate state directly. Use setState()           react/no-direct-mutation-state
```

src/components/buttons/button-accessibility-image-alt.jsx
```diff
@@ -23,19 +23,22 @@ class ButtonAccessibilityImageAlt extends React.Component {
...
+			selection.getSelectedElement() || selection.getStartElement();
+
+		this._element = element;
+
+		const imageElement = this._element.findOne('img');
 
 		const imageAlt = imageElement
 			? imageElement.getAttribute('alt')
-			: element && element.getAttribute('alt');
+			: this._element && this._element.getAttribute('alt');
```
wincent: `this._element &&` seems superflous here. If it wasn't truthy, line 35 would have thrown an error when you do `this._element.findOne('img')`.

PR #1391, opened 2020-04-10
fix: select new cell after the new row/column is inserted

SerheyDolgushev:
After the user adds a new row into the table, it is added, but the old row/cell is focused. And the toolbar is shown for the old one.
This change fixes it: after the new row is added to the table, it is selected and its toolbar is shown.

### DIFF
--- a/src/plugins/tabletools.js
+++ b/src/plugins/tabletools.js
@@ -129,7 +129,9 @@ if (!CKEDITOR.plugins.get('ae_tabletools')) {
 		return null;
 	}
 
-	function insertRow(selection, insertBefore) {
+	function insertRow(editor, insertBefore) {
+		const selection = editor.getSelection();
+
 		const cells = getSelectedCells(selection);
 
 		const firstCell = cells[0];
@@ -188,6 +190,11 @@ if (!CKEDITOR.plugins.get('ae_tabletools')) {
 		} else {
 			newRow.insertAfter(row);
 		}
+
+		const cell = new CKEDITOR.dom.element(newRow.$).getChild(
+			cells[0] ? cells[0].$.cellIndex : 0
+		);
+		selectElement(editor, cell);
 	}
 
 	function deleteRows(selectionOrRow) {
@@ -299,7 +306,9 @@ if (!CKEDITOR.plugins.get('ae_tabletools')) {
 		return retval;
 	}
 
-	function insertColumn(selection, insertBefore) {
+	function insertColumn(editor, insertBefore) {
+		const selection = editor.getSelection();
+
 		const cells = getSelectedCells(selection);
 
 		const firstCell = cells[0];
@@ -328,6 +337,7 @@ if (!CKEDITOR.plugins.get('ae_tabletools')) {
 			nextCol.push(nextCell);
 		}
 
+		const insertedCells = [];
 		for (let i = 0; i < height; i++) {
 			let cell;
 
@@ -348,8 +358,22 @@ if (!CKEDITOR.plugins.get('ae_tabletools')) {
 				cell = cell.$;
 			}
 
+			insertedCells[i] = cell;
+
 			i += cell.rowSpan - 1;
 		}
+
+		const cell = new CKEDITOR.dom.element(
+			insertedCells[firstCell.getParent().$.rowIndex]
+		);
+		selectElement(editor, cell);
+	}
+
+	function selectElement(editor, element) {
+		const range = editor.createRange();
+
+		range.moveToPosition(element, CKEDITOR.POSITION_AFTER_START);
+		editor.getSelection().selectRanges([range]);
 	}
 
 	function deleteColumns(selectionOrCell) {
@@ -851,8 +875,7 @@ if (!CKEDITOR.plugins.get('ae_tabletools')) {
 				createDef({
 					requiredContent: 'table',
 					exec(editor) {
-						const selection = editor.getSelection();
-						insertRow(selection, true);
+						insertRow(editor, true);
 					},
 				})
 			);
@@ -862,8 +885,7 @@ if (!CKEDITOR.plugins.get('ae_tabletools')) {
 				createDef({
 					requiredContent: 'table',
 					exec(editor) {
-						const selection = editor.getSelection();
-						insertRow(selection);
+						insertRow(editor);
 					},
 				})
 			);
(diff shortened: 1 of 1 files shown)


wincent:
Haven't done a proper review of this yet, but just to help you parse the CI output: amid all the warnings (which you can ignore) is one actual error in `src/plugins/tabletools.js`:

```
   202:5  error    Expected property shorthand                                                                        object-shorthand
```

The semantic pull request bot is complaining because the commit title and PR title should both start with a "type"; eg. ("fix: description of fix")

    fix: select new row after inserting in table

To provide some context with reviewing, validating this, @SerheyDolgushev, can you share with us which browser(s) you were seeing this problem on?

SerheyDolgushev:
Hello @wincent ,
I`m not really a JavaScript developer, so I was not sure about the protocols of how to test this PR. I`ve made a similar change in one of our [eZ Platform](https://github.com/ezsystems/ezplatform) project and decided to share it here too.

To test my change I did the following steps:
1. Checked out `select_inserted_row` branch
2. Run `yarn install` to get all the dependencies
3. Run `npm run build` to build
4. Set up new https://alloy-editor.localhost/ local virtual host which is pointed to `~/Projects/alloy-editor/dist`
5. Opened https://alloy-editor.localhost/index.html in Google Chrome, Version 80.0.3987.149 (Official Build) (64-bit) and run the following test:
5.1. Selected table row and click on "Insert Row After" 
add-table-row
5.2. The new row was inserted and its first cell was selected
add-table-row-2
5.3. After I typed some text it was placed in the first cell of an inserted row
add-table-row-3

I tried to run the same tests for `master` branch, and after the new row is inserted the current one is still active. And if you try to type some text it is inserted in the current table cell (the one which was active when the new row was inserting):
add-table-row-current

Please let me know if I misunderstood some of the steps, or if I should provide anything else.

SerheyDolgushev:
@wincent Please let me know if anything needs to be updated/changed here.

wincent:
So I tested this out and it seems to work pretty well. One thing that I found a bit jarring is that the selection moves unconditionally into the first column in the newly created row; I think I would have expected it to stay in the same column when entering the newly created row.

The other obvious thing that comes up when testing this is that the behavior of "Insert Column After" and "Insert Column Before" now feels a bit inconsistent, because it doesn't have the same behavior that you've added here. I don't think I'd consider merging this until we make the behavior uniform.

@julien: Would you agree that it is reasonable to do the same for columns? I'd also appreciate your thoughts on the relative risk associated with making a change like this. I am always a bit scared to touch anything in the innards of AlloyEditor... 😱

One final thought — a possible refactoring to the code you have so far:

```diff
diff --git a/src/plugins/tabletools.js b/src/plugins/tabletools.js
index 577ad547..6e102b23 100644
--- a/src/plugins/tabletools.js
+++ b/src/plugins/tabletools.js
@@ -129,7 +129,9 @@ if (!CKEDITOR.plugins.get('ae_tabletools')) {
 		return null;
 	}
 
-	function insertRow(selection, insertBefore) {
+	function insertRow(editor, insertBefore) {
+		const selection = editor.getSelection();
+
 		const cells = getSelectedCells(selection);
 
 		const firstCell = cells[0];
@@ -189,11 +191,7 @@ if (!CKEDITOR.plugins.get('ae_tabletools')) {
 			newRow.insertAfter(row);
 		}
 
-		return new CKEDITOR.dom.element(newRow.$);
-	}
-
-	function selectRowsCell(row, selection, editor) {
-		const cell = row.getChild(0);
+		const cell = new CKEDITOR.dom.element(newRow.$).getChild(0);
 
 		selection.selectElement(cell);
 
@@ -867,9 +865,7 @@ if (!CKEDITOR.plugins.get('ae_tabletools')) {
 				createDef({
 					requiredContent: 'table',
 					exec(editor) {
-						const selection = editor.getSelection();
-						const row = insertRow(selection, true);
-						selectRowsCell(row, selection, editor);
+						insertRow(editor, true);
 					},
 				})
 			);
@@ -879,9 +875,7 @@ if (!CKEDITOR.plugins.get('ae_tabletools')) {
 				createDef({
 					requiredContent: 'table',
 					exec(editor) {
-						const selection = editor.getSelection();
-						const row = insertRow(selection);
-						selectRowsCell(row, selection, editor);
+						insertRow(editor);
 					},
 				})
 			);
```

`insertRow` is a private function, so it seems an improvement to make updating the selection an integral part of insertion rather than calling it separately afterwards.

julien:
@wincent just so you know I can confirm the behaviour looks correct **but** I agree that 
1. The fact the selection always moves into the first column is something that should be improved 
2. With this change, the "Insert Column After" and "Insert Column Before" behaviour are different and it does feel inconsistent

I also agree that it would be reasonable to do this for columns.

As for relative risks associated with this change I'd say that on one hand it's an improvement and that's always good, but on the other hand, this is not the same behaviour as CKEditor as seen in the following gif . So while I don't see any major risk, I think that as you mentioned it can feel inconsitent and that's also something we want to avoid   

ckeditor

wincent:
> on the other hand, this is not the same behaviour as CKEditor as seen in the following gif

Wow, CKEditor in that GIF looks messed up.

- It looks like you selected "Insert Row Before", but CKEditor inserted the row after (and the cursor stayed in the original cell).
- Then you selected "Insert Row After" and it appeared to insert the row after, but maybe it inserted it before (not clear because this time the cursor *did* move, so I don't know which one was the original cell).

SerheyDolgushev:
Sorry, for so long delay. I'm going to have a look at this one this week.

SerheyDolgushev:
The following changes have been made:
1. The same cell is selected after new row is inserted
2. The same cell is selected after new column is inserted
3. Selecting the cell means moving the cursor in it, instead of selecting its empty content

@wincent @julien please review those changes.

I`m not sure why Travis check failed, do not think it is related to the latest changes. Also please provide the semantically valid PR title/commit.

I will squash all the commits from this PR if everything looks good for you gentlemen.

wincent:
Thanks.

> Also please provide the semantically valid PR title/commit.

That just means that the PR title *and* every commit need a valid "type" (in this case, that means they have to start with "fix:").

> I`m not sure why Travis check failed

Try running `yarn format` — I think it is probably complaining about formatting. If you want to do that, rebase/squash everything into one commit, update the commit title to start with "fix:" and force-push, we can do a final test and take a final look at this.

SerheyDolgushev:
@wincent please have a look

SerheyDolgushev:
`yarn format` indeed fixed the Travis issue. Thanks! :)

wincent:
Looks good to me @SerheyDolgushev.

PR #1394, opened 2020-05-18
fix: gracefully handle invalid JSON in data-style attribute (#1393)

### LINKED ISSUE #1393: Can't embed youtube video because of wrong attribute data-styles="{"width":"100%"}"
**Do you want to request a _feature_ or report a _bug_?**
Report a bug.

**What is the current behaviour?**
When you try to embed  a youtube video using the following code: "<div class="embed-responsive embed-responsive-16by9" data-embed-id="Qv1nRlct5a4" data-styles="{"width":"100%"}" style="width:100%"><iframe allow="autoplay; encrypted-media" allowfullscreen="" frameborder="0" height="315" src="https://www.youtube.com/embed/Qv1nRlct5a4?rel=0" tabindex="0" width="560"></iframe></div>"
In the console an exception is being thrown"VM5500:1 Uncaught SyntaxError: Unexpected token w in JSON at position 1
    at JSON.parse (<anonymous>)
    at h.data (embedurl.js:413)
    at h.g (ckeditor.js?browserId=other&languageId=en_US&b=7201&t=1588939054524:10)
    at h.fire (ckeditor.js?browserId=other&languageId=en_US&b=7201&t=1588939054524:11)
    at new h (plugin.js?t=K3FH:7)
    at q.initOn (plugin.js?t=K3FH:55)
    at q.initOnAll (plugin.js?t=K3FH:55)
    at a.<anonymous> (plugin.js?t=K3FH:28)
    at a.g (ckeditor.js?browserId=other&languageId=en_US&b=7201&t=1588939054524:10)
    at a.<anonymous> (ckeditor.js?browserId=other&languageId=en_US&b=7201&t=1588939054524:11)"
This makes possible to save a wrong HTML code without any possibility of edit or remove it after it. 

**If the current behaviour is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
1. Try to embed the following code: "<div class="embed-responsive embed-responsive-16by9" data-embed-id="Qv1nRlct5a4" data-styles="{"width":"100%"}" style="width:100%"><iframe allow="autoplay; encrypted-media" allowfullscreen="" frameborder="0" height="315" src="https://www.youtube.com/embed/Qv1nRlct5a4?rel=0" tabindex="0" width="560"></iframe></div>"
2. An exception is thrown at "embedurl.js:413"
3. You can save wrong styles text and without any possible way of editing or removing after it. 

**What is the expected behaviour?**
You can save correct style text and with the possibility of editing or removing after it. 

Affects version 2.11.6, affects all browsers. The bug begins when embedurl.js was created and added.

georgel-pop-lr:
Is fixing #1393 If data-style attribute is not a valid JSON an exception will be thrown and the code will stop working.

### DIFF
--- a/src/plugins/embedurl.js
+++ b/src/plugins/embedurl.js
@@ -406,11 +406,19 @@ if (!CKEDITOR.plugins.get('embedurl')) {
 
 					// Sync dimensions and alignment with editor wrapper
 
+					let styles = null;
+
 					const stylesJSON = instance.element.getAttribute(
 						'data-styles'
 					);
 
-					let styles = stylesJSON ? JSON.parse(stylesJSON) : null;
+					if (stylesJSON) {
+						try {
+							styles = JSON.parse(stylesJSON);
+						} catch (_error) {
+							styles = null;
+						}
+					}
 
 					if (!styles) {
 						const iframe = instance.wrapper.findOne('iframe');


wincent:
Thanks @georgel-pop-lr. Can you please run `yarn format` on this (that's what CI is complaining about) and do a `git commit --amend` and force push to update this with a new title (both the commit subject line and the PR title need to start with "fix: " otherwise the semantic PR bot will complain (not sure why it hasn't yet). You could use a subject/title like this:

    fix: gracefully handle invalid JSON in data-style attribute (#1393)

georgel-pop-lr:
> Thanks @georgel-pop-lr. Can you please run `yarn format` on this (that's what CI is complaining about) and do a `git commit --amend` and force push to update this with a new title (both the commit subject line and the PR title need to start with "fix: " otherwise the semantic PR bot will complain (not sure why it hasn't yet). You could use a subject/title like this:
> 
> ```
> fix: gracefully handle invalid JSON in data-style attribute (#1393)
> ```

Thanks @wincent, hope is fine now.

src/plugins/embedurl.js
```diff
@@ -406,11 +406,21 @@ if (!CKEDITOR.plugins.get('embedurl')) {
...
 					);
 
-					let styles = stylesJSON ? JSON.parse(stylesJSON) : null;
+					if (stylesJSON) {
+						try {
+							styles = JSON.parse(stylesJSON);
+						} catch (e) {
+							console.warn(
+								'Unable to parse data-styles attribute!'
+							);
```
wincent, line 421: Couple things here:

1. Linter may warn about unused variable `e` here, so either use it (in the log statement) or mark it as intentionally unused by starting it with an underscore (ie. `_error`). Note also that we try to avoid abbreviations if we can (ie. prefer `error` over `e`), except for things like `i` for loop variables.
2. We generally try to avoid anything that might resembele console spew in Liferay projects because it can be hard for customers to distinguish between bugs and mere diagnostic information (see [guidelines on this](https://github.com/liferay/liferay-frontend-guidelines/blob/master/dxp/debugging.md)); sometimes they see console output as a sign of bugginess and any "helpful" output we might include actually just confuses them. And they may be alarmed to see a message ending in an exclamation point. As such, I'd actually just swallow this error rather than log it — in some browsers, the stack trace that comes with `console.warn` looks particularly scary.

PR #1404, opened 2020-06-29
feat: make the toolbar visible in tables

### LINKED ISSUE #1403: Table toolbar is not visible in case of many rows
The table toolbar is not visible if the top of the table is not visible when a table is created with many rows.

Reproduction steps:

Make a table that has 50 rows and 3 columns
Scroll down to the last row of the table and click on it
Expected:
The Table toolbar to add/remove rows/columns/cells to appear top-center of the cell

Actual behavior
The toolbar appears, but it's not visible. In fact, it will be at the very top-center of the table

It is reproducable on the AlloyEditor demo @ https://alloyeditor.com/

The customer ticket: https://liferay-support.zendesk.com/agent/tickets/26861
The TSE ticket: https://issues.liferay.com/browse/LPP-37979
My investigatiion: https://issues.liferay.com/browse/LPP-37979?focusedCommentId=2188689&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2188689

julien:
This change, makes the toolbar always visible when editing tables,
The toolbar will move above the current row or column currently being
edited.

Closes #1403

### DIFF
--- a/src/selections/selection-position.js
+++ b/src/selections/selection-position.js
@@ -46,9 +46,23 @@ const centerToolbar = function(toolbar, rect) {
 		CKEDITOR.SELECTION_BOTTOM_TO_TOP
 	);
 
+	const caretPosition = nativeEditor.getSelection();
+	const ranges = caretPosition.getRanges();
+	let offsetHeight = 0;
+
+	if (ranges) {
+		const startContainer = ranges[0].startContainer;
+		const startContainerClientRect = startContainer.getClientRect();
+		offsetHeight = startContainerClientRect.y - rect.top;
+	}
+
 	const endPosition = [
 		rect.left + rect.width / 2 - halfNodeWidth - scrollPosition.x,
-		rect.top - toolbarNode.offsetHeight + scrollPosition.y - gutter.top,
+		rect.top +
+			offsetHeight -
+			toolbarNode.offsetHeight +
+			scrollPosition.y -
+			gutter.top,
 	];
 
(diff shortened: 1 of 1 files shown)


wincent:
I'm not sure that this is the right fix, from a UX perspective, because it makes it seem like the buttons in the toolbar apply to the specific cell that the toolbar is floating next to.

Not such a big deal for things like "Insert Column After", where the item is very much positional:

Screenshot 2020-06-30 -121751-9bwW8E89@2x

But it would be a bad thing for the "Delete" button; the positioning of the toolbar makes it look like it might delete the contents of a cell, and not the entire table (it will delete the entire table).

I am wondering if it would be better to instead scroll the toolbar into view on creating a large table. I know you implemented the "expected behavior" from #1403:

> The Table toolbar to add/remove rows/columns/cells to appear top-center of the cell

But I would like to get some input from UX to make sure we're on the right track here.

wincent:
Legally binding and highly rigorous UX review [here](https://liferay.slack.com/archives/G3JBR21HA/p1593512483254700):

> fix is valid because it adds better understanding for the context actions (format headings and delete table are actions no aware of the context).
>
> it may cause an unexpected behavior with the delete table action. But undo action is available (via keyboard) so no big deal for those users.
>
> In addition, using a different icon (table + times icon) could prevent this behavior too. Even a dropdown button with all the delete actions for column, row, table.

So going to merge.

launch

PR #1409, opened 2020-07-20
fix: Don't allow users to input a value less than a minimum `row` or `col` value when creating a table

diegonvs:
…`col` value when creating a table

### Test Case:

Run Demo Page:

Click on AE's plus Icon, add a Table with 0 for number of columns and -1 for number of rows

Currently AE allows the user for input negative values for cols/rows.

### What I did:

I restricted the user to set a value less than the minimum value for col/row defined on html element just preventing the user to type negative values from the keyboard. That was already defined to `-1`. See https://github.com/liferay/alloy-editor/compare/master...diegonvs:110423?expand=1&title=fix%3A+Don%27t+allow+users+to+input+a+value+less+than+a+minimum+%60row%60+or+%60col%60+value+when+creating+a+table#diff-4b5fe36c64a18eb8591a90632c092ee0L164

@drakonux, from a UX perspective, is it correct? An alternative can be providing a validation message.

Closes: https://issues.liferay.com/browse/LPS-110423

### DIFF
--- a/src/components/buttons/button-table-edit.jsx
+++ b/src/components/buttons/button-table-edit.jsx
@@ -121,12 +121,8 @@ class ButtonTableEdit extends React.Component {
 	_handleChange = (inputName, event) => {
 		const state = {};
 
-		if (
-			inputName === INPUT_NAMES.ROWS ||
-			(inputName === INPUT_NAMES.COLS &&
-				event.target.value < MINIMUM_GRID_VALUE)
-		) {
-			state[inputName] = MINIMUM_GRID_VALUE;
+		if (inputName === INPUT_NAMES.COLS || inputName === INPUT_NAMES.ROWS) {
+			state[inputName] = Math.min(event.target.value, MINIMUM_GRID_VALUE);
 		} else {
 			state[inputName] = event.target.value;
 		}
--- a/src/components/buttons/button-table-edit.jsx
+++ b/src/components/buttons/button-table-edit.jsx
@@ -11,6 +11,13 @@ import ButtonIcon from './button-icon.jsx';
 const KEY_ENTER = 13;
 const KEY_ESC = 27;
 
+const INPUT_NAMES = {
+	COLS: 'cols',
+	ROWS: 'rows',
+};
+
+const MINIMUM_GRID_VALUE = 1;
+
 /**
  * The ButtonTableEdit class provides functionality for creating and editing a table in a document.
  * Provides UI for creating a table.
@@ -113,7 +120,16 @@ class ButtonTableEdit extends React.Component {
 	 */
 	_handleChange = (inputName, event) => {
 		const state = {};
-		state[inputName] = event.target.value;
+
+		if (
+			inputName === INPUT_NAMES.ROWS ||
+			(inputName === INPUT_NAMES.COLS &&
+				event.target.value < MINIMUM_GRID_VALUE)
+		) {
+			state[inputName] = MINIMUM_GRID_VALUE;
+		} else {
+			state[inputName] = event.target.value;
+		}
 
 		this.setState(state);
 	};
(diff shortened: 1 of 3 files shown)


drakonux:
Very interesting question. 

A machine/system/UI should not change the users' decision (a user must able to type 0 rows or column) in fact as far as I know the [HTML5 input component](https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_input_max_min) with the min or max attributes set. It doesn't change the value when it's below or above, and a validation tooltip is shown.

However, we could subscribe to make an exception here. Because designing a good behavior for interfaces has also to prevent users from making mistakes and it makes no sense to allow users to type 0 for creating a table. So by default when a user wants to create a table the default and minimum value for row and column must be 1. 

Anyway, as a further fix. I think we could provide a screen reader message when the interfaces change the value to 1 like "Value automatically adjusted to 1" or something similar, what do you think? @marcoscv-work

duracell80:
IMG_20200721_092718_copy_1512x2016

Guys don't overthink it. CKEditor doesn't allow 0 or minus 1 and throws a simple alert message when entering anything less than 1. Since the base editor doesn't allow the user to enter 0 ... Why should Alloy?

carloslancha:
Hey @duracell80 I think that's exactly what @drakonux is talking about 😉 

> So by default when a user wants to create a table the default and minimum value for row and column must be 1 (edited - so the user can't set a value below 1).
Anyway, as a further fix. I think we could provide a screen reader message when the input changes the value to 1 like "Value automatically adjusted to 1" or something similar

src/components/buttons/button-table-edit.jsx
```diff
@@ -113,7 +120,16 @@ class ButtonTableEdit extends React.Component {
 	 */
 	_handleChange = (inputName, event) => {
 		const state = {};
-		state[inputName] = event.target.value;
+
+		if (
+			inputName === INPUT_NAMES.ROWS ||
+			(inputName === INPUT_NAMES.COLS &&
+				event.target.value < MINIMUM_GRID_VALUE)
```
wincent, line 127: I don't understand this conditional. It says:

- If input is "rows", set value to 1.
- If input is "cols" and value is less than 1, set value to 1.

Maybe you the parentheses are in the wrong place and you meant:

- If input is "rows" and value is less than 1, set value to 1.
- If input is "cols" and value is less than 1, set value to 1.

ie.

```js
if (
  (inputName === INPUT_NAMES.COLS || inputName === INPUT_NAMES.ROWS) &&
  event.target.value < MINIMUM_GRID_VALUE
)
```

Having said that, a common pattern for that kind of thing is just to use `Math.min()`:

```js
if (inputName === INPUT_NAMES.COLS || inputName === INPUT_NAMES.ROWS) {
    state[inputName] = Math.min(event.target.value, MINIMUM_GRID_VALUE);
}
```

test/ui/test/button-table-edit.jsx
```diff
@@ -13,6 +13,34 @@ describe('ButtonTableEdit', function() {
 	beforeEach(Utils.createAlloyEditor);
 	afterEach(Utils.destroyAlloyEditor);
 
+	it('should create a 1x1 table by default when user selects a negative value for rows or columns', function() {
```
wincent, line 16: As per [`liferay/no-it-should` ESLint rule](https://github.com/liferay/eslint-config-liferay/blob/HEAD/plugins/eslint-plugin-liferay/docs/rules/no-it-should.md), I'd suggest changing this from "should create" to "creates".

marcoscv-work:
Yeah! about error prevention:
https://www.w3.org/TR/UNDERSTANDING-WCAG20/minimize-error.html
https://www.w3.org/TR/UNDERSTANDING-WCAG20/minimize-error-reversible.html

src/components/buttons/button-table-edit.jsx
diegonvs:
```javascript

test/ui/test/button-table-edit.jsx
diegonvs: updated whole file

PR #1411, opened 2020-07-22
chore: Updated CKEditor build to v4.14.1 in AlloyEditor 1.5

dew326:
As I'm not able to update my AlloyEditor to version 2.x I updated the CKEditor build to v4.14.1 for the AlloyEditor 1.5.
I used your config and downloaded the "optimized" version of CKEditor.

@wincent Is there any chance this could be merged and released as the next patch release of the 1.5 version?


wincent:
> @wincent Is there any chance this could be merged and released as the next patch release of the 1.5 version?

Yes, but we need to do a few things:

- See a green CI run.
- Sanity check by looking over the corresponding updates we made on 2.x (ie. the `master` branch) to see if there were any additionally tweaks we had to make for compatibility with the newer version of CKEditor (off the top of my head, I don't recall).
- Understand what your motivation of moving to the new version, because this is a big change which obviously entrails risk, and given that 1.x is in maintenance mode, we have to be very careful about regressions lest it become an unstable time sink.

dew326:
My motivation is to fix the security issue https://nvd.nist.gov/vuln/detail/CVE-2020-9281. 
In my project, I cannot update to 2.x because of the BC breaks. 
I tried to use the version without CKEditor and load it manually on my page but this didn't work at all. For some unknown reason, it started to use the `create` function here (https://github.com/liferay/alloy-editor/blob/12a8ff13f3cfad71326613184f2003ab77e7c7dd/src/components/uibridge/button.jsx#L154) and when invoking the `generateButtonBridge()` there are only two arguments and in the definition of this function are three parameters (https://github.com/liferay/alloy-editor/blob/12a8ff13f3cfad71326613184f2003ab77e7c7dd/src/components/uibridge/button.jsx#L25) that's why I had `editor` undefined. When I fixed those issues I was missing `this.props` in the `render` method.
So I gave up on this method of solving the issue. Do you know if did somethings work or there are some bugs?

For the CI, I think the problem is that there is missing the Promise polyfill (https://travis-ci.org/github/liferay/alloy-editor/builds/710689107#L1214) but I added this here https://github.com/liferay/alloy-editor/pull/1411/files#diff-5476759aef8ec7318cd1971d24a709d2R1 and when I do `yarn prepublishOnly` it is successfully copied to the dist catalog, do you release in another way on the CI environment?

wincent:
> My motivation is to fix the security issue https://nvd.nist.gov/vuln/detail/CVE-2020-9281.
> In my project, I cannot update to 2.x because of the BC breaks.

Thanks for the context; that helps.

> So I gave up on this method of solving the issue. Do you know if did somethings work or there are some bugs?

Off the top of my head, not really. 1.x [hasn't really been touched](https://github.com/liferay/alloy-editor/commits/1.x) since about a year ago, and even then it was mostly pretty minimal bug-fixing only; my memory of what the status of it is is pretty foggy.

> For the CI, I think the problem is that there is missing the Promise polyfill

In addition, it could just be flaky. Generally in those multi-browser tests, it's not surprising to see failures on IE or Edge... Sometimes the bug is legit, but other times it goes away with a re-run.

> when I do `yarn prepublishOnly` it is successfully copied to the dist catalog, do you release in another way on the CI environment?

In CI [it's running](https://github.com/liferay/alloy-editor/blob/1.x/.travis.yml) `yarn testSaucelabs`.

In `prepublishOnly` [it just runs](https://github.com/liferay/alloy-editor/blob/1.x/package.json#L58) `gulp clean && gulp release`.

dew326:
I solved the issue with missing `promise.js` and now it has some random issues on Edge (each time it has different errors) so I assume those are some false failures. 
Should I re-run the tests until all are green?

dew326:
All green 🎉

wincent:
In terms of licensing, where did `lib/ckeditor/vendor/promise.js` come from?

dew326:
This comes from the CKEditor package, I used the link for download the latest version of optimized CKEditor in build-config.js (http://ckeditor.com/builder/download/e05e62fb481ae83a4074f743c787736f)

wincent:
@dew326: I pushed [one more commit](https://github.com/liferay/alloy-editor/pull/1411/commits/9bfbb818629bc90ecf1cfff336ba5707605fbbbe) on top of your branch.

The basic idea is to copy [the build script that we have on `master`](https://github.com/liferay/alloy-editor/blob/185f441f12aef5df6325c354db38b32b0dfc5e4a/scripts/build/build-ckeditor.sh) and use that to create the build with `scripts/build/build-ckeditor.sh 4.14.1`.

- The CKEditor source lives in a submodule at `lib/ckeditor-dev`.
- The script checks out the requested version, runs the build etc, updates the artifacts in `lib/ckeditor`.

This means that we don't have to review the (almost) binary blobs in a PR like this one, because we know they came from the upstream source (ie. we trust the person running the script, and even if we don't trust the person, we can just re-run the script and verify that we get the same result).

In practical terms, the only difference between your change and what the build produces are line-endings (due to quirkiness of the build process) and there are some minor content differences because some things come differently with the zip compared to the build (eg. the README.md).

We're using exactly the same build-config file as in `master` (note the tiny diff there: the addition of the 'pastetools' plugin — master commit for that one was https://github.com/liferay/alloy-editor/commit/62ddf171f1a7d95c6da9376580a1fa4e7af6d704).

Let's see if CI likes this one... 🤞

wincent:
> Let's see if CI likes this one... 🤞

And of course it doesn't 😂

```
Edge 14.14393.0 (Windows 10.0.0) ERROR
  extend failed, verify dependencies
```

Will retry.

wincent:
Ok, I'm going to leave this open overnight to give people a chance to comment on it, but otherwise will just merge it (probably tomorrow) and we can cut a release after that.

wincent:
> I'm going to leave this open overnight to give people a chance to comment on it, but otherwise will just merge it

And As It Is Such, So Also As Such Is It Unto You... ie. merging.

dew326:
Thank you.

PR #1414, opened 2020-07-23
fix: unbreak table drag-resizing by using fixed table layout

diegonvs:
When inserting data, everything must be wrapped up on the cell and cell width couldn't expand.

It is impossible to use the drag ability to resize a column when other columns in the table have content which then pushes the empty columns together.

### How I fixed it: 

Based on https://github.com/ckeditor/ckeditor4/issues/3614#issuecomment-549410457 I just changed the table style directly on the table definition. 

### Before:
fail

### After:
mammamia

#### Resize:

resize

Fixes: https://issues.liferay.com/browse/LPS-114429

### DIFF
--- a/src/components/buttons/button-table-edit.jsx
+++ b/src/components/buttons/button-table-edit.jsx
@@ -32,7 +32,7 @@ class ButtonTableEdit extends React.Component {
 			border: 1,
 			cellPadding: 0,
 			cellSpacing: 0,
-			style: 'width: 100%',
+			style: 'table-layout: fixed; width: 100%;',
 		},
 	};
 


wincent:
This fix looks reasonable to me, but it is not without its risk because it is going to affect all AlloyEditor instances by default.

When I search for `tableAttributes` in liferay-portal, I find [one ConfigContributor](https://github.com/liferay/liferay-portal/blob/d0e51c547f753dda645d703007c228dff7edd49f/modules/apps/frontend-editor/frontend-editor-alloyeditor-web/src/main/java/com/liferay/frontend/editor/alloyeditor/web/internal/editor/configuration/AlloyEditorCreoleConfigContributor.java#L191) that overriding this, but none of [the other ConfigContributors](https://github.com/liferay/liferay-portal/tree/d0e51c547f753dda645d703007c228dff7edd49f/modules/apps/frontend-editor/frontend-editor-alloyeditor-web/src/main/java/com/liferay/frontend/editor/alloyeditor/web/internal/editor/configuration) do that...

Having said that, current behavior seems so unpleasant that it is probably worth the risk...

One other thing; in testing this, I think I found a regression introduced by bef3ce28:

Screenshot 2020-07-24 -124412-ZHLamCxs@2x

Reverting that commit, errors go away, and table toolbar is visible. Not sure why this didn't come up in testing of #1404 — I think it is likely an interaction with the subsequent CKEditor update. I'm going to take a quick look into that now, but it's unrelated to this PR so you can ignore it.

wincent:
I added fix for the unrelated JS error [here](https://github.com/liferay/alloy-editor/pull/1415).

wincent:
What could go wrong? Going to merge.

what-could-go-wrong

jbalsas:
whynot

PR #1419, opened 2020-08-10
fix: Avoids blocking user input when settling a minimal input value

diegonvs:
### Test Case

1. Open "Advanced Editing Demo" from the current master
2. Click on the add button for adding a grid
3. Try to type the number `22` cleaning the current value and just typing `22`

It's not possible settling this number because the onChange event is being triggered for every keyboard input like `delete`.

before

### Behavior changed

I tried to fix it changing from `onChange` to `onInput` but I didn't get success, also, I tried to use `onBlur` for just updating the input when the input is blurred but no success. The only alternative that I've found was when submitting the value from the inputs, validate, and then just changing the minimum value if received an invalid value.

After the mentioned changes:

after

### DIFF
--- a/src/components/buttons/button-table-edit.jsx
+++ b/src/components/buttons/button-table-edit.jsx
@@ -100,8 +100,8 @@ class ButtonTableEdit extends React.Component {
 
 		tableUtils.create({
 			attrs: this.props.tableAttributes,
-			cols: cols < MINIMUM_GRID_VALUE ? MINIMUM_GRID_VALUE : cols,
-			rows: rows < MINIMUM_GRID_VALUE ? MINIMUM_GRID_VALUE : rows,
+			cols: Math.max(MINIMUM_GRID_VALUE, cols),
+			rows: Math.max(MINIMUM_GRID_VALUE, rows),
 		});
 
 		this.props.cancelExclusive();
--- a/src/components/buttons/button-table-edit.jsx
+++ b/src/components/buttons/button-table-edit.jsx
@@ -96,11 +96,12 @@ class ButtonTableEdit extends React.Component {
 	_createTable = () => {
 		const editor = this.context.editor.get('nativeEditor');
 		const tableUtils = new CKEDITOR.Table(editor);
+		const {cols, rows} = this.state;
 
 		tableUtils.create({
 			attrs: this.props.tableAttributes,
-			cols: this.state.cols,
-			rows: this.state.rows,
+			cols: cols < MINIMUM_GRID_VALUE ? MINIMUM_GRID_VALUE : cols,
+			rows: rows < MINIMUM_GRID_VALUE ? MINIMUM_GRID_VALUE : rows,
 		});
 
 		this.props.cancelExclusive();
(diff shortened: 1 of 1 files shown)


src/components/buttons/button-table-edit.jsx
```diff
@@ -96,11 +96,12 @@ class ButtonTableEdit extends React.Component {
...
 		const editor = this.context.editor.get('nativeEditor');
 		const tableUtils = new CKEDITOR.Table(editor);
+		const {cols, rows} = this.state;
 
 		tableUtils.create({
 			attrs: this.props.tableAttributes,
-			cols: this.state.cols,
-			rows: this.state.rows,
+			cols: cols < MINIMUM_GRID_VALUE ? MINIMUM_GRID_VALUE : cols,
+			rows: rows < MINIMUM_GRID_VALUE ? MINIMUM_GRID_VALUE : rows,
```
wincent, line 104: Seeing as you used a `Math` function for this before (ie. on the deleted line 125) you should probably use it here too.

But note, the old version was wrong (probably because you used the _untested snippet_ that I suggested [here](https://github.com/liferay/alloy-editor/pull/1409#discussion_r459895177)). Please remember that suggestions like that are often just typed into a text field in a browser on GitHub and almost certainly haven't been tested beforehand; you need to critically evaluate each suggestion and validate that it actual does what it is supposed. Should have been `Math.max`, not `Math.min`.

```js
cols: Math.max(MINIMUM_GRID_VALUE, cols),
rows: Math.max(MINIMUM_GRID_VALUE, rows),
```

wincent: ☝️ that one is also an untested snippet.

diegonvs: I tested this snippet and looks fine ;) Updated!

src/components/buttons/button-table-edit.jsx
```diff
@@ -121,11 +122,7 @@ class ButtonTableEdit extends React.Component {
 	_handleChange = (inputName, event) => {
 		const state = {};
 
-		if (inputName === INPUT_NAMES.COLS || inputName === INPUT_NAMES.ROWS) {
-			state[inputName] = Math.min(event.target.value, MINIMUM_GRID_VALUE);
-		} else {
-			state[inputName] = event.target.value;
-		}
+		state[inputName] = event.target.value;
```
wincent, line 125: Instead of:

```js
const state = {};

state[inputName] = event.target.value;

this.setState(state);
```

this could be:

```js
this.setState({
    [inputName]: event.target.value,
});
```

but funnily enough, your result is almost the same [as it used to be](https://github.com/liferay/alloy-editor/blob/a1532cac870dd25f58e4f8db89d2b4f7bac399d3/src/components/buttons/button-table-edit.jsx#L115-L118) before all the recent changes:

```js
const state = {};
state[inputName] = event.target.value;

this.setState(state);
```

wincent: Added: https://github.com/liferay/alloy-editor/pull/1419#discussion_r469162192

wincent:
@diegonvs: I force-pushed to your fork to update this:

- Retitled [second commit](https://github.com/liferay/alloy-editor/pull/1419/commits/71aeed95cc97332f7fd074666a510c6586a8aec2) as "refactor:" instead of "fix:" because it was a refactor commit.
- [Added simplification](https://github.com/liferay/alloy-editor/pull/1419/commits/6a15a9f822889a0b695c53359af61020c83fad1b) as suggested [here](https://github.com/liferay/alloy-editor/pull/1419#discussion_r469162192).

wincent:
CI is green, and I repeated your test plan @diegonvs and this looks good to merge:

Screenshot 2020-08-12 -123910-l14LvsOk@2x

wincent:
Released: https://github.com/liferay/alloy-editor/releases/tag/v2.11.12

diegonvs:
Thanks @wincent! :)

PR #1428, opened 2020-09-07
fix: use `this.props.editor` instead of `this.context.editor`

julien:
When looking at [PTR-1914](https://issues.liferay.com/browse/PTR-1914),
`this.context.editor` was `undefined` in both of these `isActive`
functions.

I also spotted this
[comment](https://github.com/liferay/alloy-editor/blob/f3617e1e629fd13553c4a933efa547198eddbb2f/src/adapter/editor-context.js#L14-L21) which makes me think that this is a 
reasonable change 
(after having tested the change locally and seeing it not failing).

### DIFF
--- a/src/components/base/button-command-active.js
+++ b/src/components/base/button-command-active.js
@@ -24,7 +24,7 @@ export default WrappedComponent =>
 		 * @return {Boolean} True if the command is active, false otherwise.
 		 */
 		isActive() {
-			const editor = this.context.editor.get('nativeEditor');
+			const editor = this.props.editor.get('nativeEditor');
 
 			const command = editor.getCommand(this.props.command);
 
--- a/src/components/base/button-style.js
+++ b/src/components/base/button-style.js
@@ -89,7 +89,7 @@ export default WrappedComponent =>
 		 * @return {Boolean} True if style is active, false otherwise.
 		 */
 		isActive() {
-			const editor = this.context.editor.get('nativeEditor');
+			const editor = this.props.editor.get('nativeEditor');
 			const elementPath = editor.elementPath();
 			return this.getStyle().checkActive(elementPath, editor);
 		}


wincent:
> (after having tested the change locally and seeing it not failing).

I'm not super confident about this change, although I believe you when you say it works... my fear is that it might only be working by accident.

If you look at [the original commit](https://github.com/liferay/alloy-editor/commit/86b5dd8959d9e4ebf4047eb19cdd6119732ed3ab) in which I added the `EditorContext.toProps` thing, you'll see that I user `EditorContext.toProps` to explicitly wrap the components that wanted to read `this.props.context` instead of `this.context`. ie.

Old code:

```js
const editor = this.context.editor.get('nativeEditor');	
// ...
export default ButtonEmbedEdit;
```

New code, with wrapping:

```js
const editor = this.props.context.editor.get('nativeEditor');
// ...
export default EditorContext.toProps(ButtonEmbedVideoEdit);
```

So, in this PR, there is no wrapping, yet "somehow" the context is getting passed in via props. That's what makes me feel like this might be working _by accident_.

As noted in 86b5dd8959d9e4ebf4047eb19cdd6119732ed3ab, the behavior of the legacy context API is dependent on the version of React you're talking about, and if you look at the [current docs for it](https://reactjs.org/docs/legacy-context.html#referencing-context-in-lifecycle-methods), it makes it seem like it has changed again since then (for example, now it says that constructors get context as a second param, and right now I can't remember if it used to be like that).

All of which is to say, is that this change might be perfectly valid, but I am nervous about it, so we should test it very thoroughly. If you're confident that it works, feel free to ship it.

julien:
@wincent after reading your comment, I also have doubts about how to solve this.

What's happening in this case is that `this.context` in the [`isActive`](https://github.com/liferay/alloy-editor/blob/a8c76130f4cea1d4ad6c3da77285b3575b74fb6b/src/components/base/button-style.js#L92) function in [`ButtonStyle.js`](https://github.com/liferay/alloy-editor/blob/a8c76130f4cea1d4ad6c3da77285b3575b74fb6b/src/components/base/button-style.js) is just a plain object with no properties



So I'll keep on digging and see what I can do about it - the fact that the behavior of the legacy context API is dependent on the version of React is a bit scary, but I'll probably discard this PR.

julien:
@wincent I also saw that in [`3c7a73656ab04a37c6587c8fa533fc9cedd98ece`](https://github.com/liferay/alloy-editor/commit/3c7a73656ab04a37c6587c8fa533fc9cedd98ece) we  did the "opposite" of what I'm changing here.

I'm going to close this for now.

julien:
I'm going to reopen this since I'm still looking at https://issues.liferay.com/browse/PTR-1914

So far I've been debugging Senna.js to check if I could spot something but no luck ... 
There's no context in "custom" button components but only in the second render

julien:
This can be closed.

PR #1429, opened 2020-09-07
chore: config Yarn to use "pre" as preid

wincent:
So that we can create prerelease test releases like we do in liferay-js-themes-toolkit with `yarn version --prepatch` and so on:

- https://classic.yarnpkg.com/en/docs/cli/version/
- https://github.com/liferay/liferay-js-themes-toolkit/blob/cefa75c9c64/.yarnrc (ie. current master)

I used this to produce

- https://github.com/liferay/alloy-editor/releases/tag/v2.11.14-pre.0

And documented the process here:

- https://github.com/liferay/alloy-editor/wiki/Contributing/_compare/7f5484e35a...28f26cca11

### DIFF
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,3 +1,4 @@
 # Make `yarn version` produce the right commit message and tag for this package.
 version-tag-prefix "v"
 version-git-message "chore: prepare v%s release"
+--version.preid pre


wincent:
No takers? I already tested this when I cut the release in the description, so may as well merge it.

julien:
I was going to have a look at it.

Since I didn't know about this I had a look at [this](https://classic.yarnpkg.com/en/docs/cli/version/#toc-yarn-version-preid)  and also https://github.com/liferay/liferay-js-themes-toolkit/blob/cefa75c9c64/.yarnrc which makes me think this is good to go.

julien (approved):

julien:
> so may as well merge it.
indeed

PR #1439, opened 2020-09-30
fix: avoid accessing this.context outside of render and lifecycle methods

NemethNorbert:
Detailed discussion that led to this change can be found here:
https://issues.liferay.com/browse/PTR-1914 .

According to the official docs this.context should only be accessed in
lifecycle methods including the render function.
https://reactjs.org/docs/context.html#classcontexttype

I also go into more detail why I decided to apply these changes here:
https://issues.liferay.com/browse/PTR-1914?focusedCommentId=2278112&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-2278112

### DIFF
--- a/src/components/base/toolbar-buttons.js
+++ b/src/components/base/toolbar-buttons.js
@@ -383,7 +383,7 @@ export default WrappedComponent =>
 		 */
 		show() {
 			const domNode = ReactDOM.findDOMNode(this);
-			const uiNode = this.context.editor.get('uiNode');
+			const uiNode = this.props.editor.get('uiNode');
 
 			const scrollTop = uiNode ? uiNode.scrollTop : 0;
 
@@ -443,7 +443,7 @@ export default WrappedComponent =>
 
 			if (interactionPoint && domNode) {
 				const uiNode =
-					this.context.editor.get('uiNode') || document.body;
+					this.props.editor.get('uiNode') || document.body;
 				const uiNodeStyle = getComputedStyle(uiNode);
 				const uiNodeMarginLeft = parseInt(
 					uiNodeStyle.getPropertyValue('margin-left'),
@@ -520,7 +520,7 @@ export default WrappedComponent =>
 		getToolbarButtons(buttons, additionalProps) {
 			const buttonProps = {};
 
-			const nativeEditor = this.context.editor.get('nativeEditor');
+			const nativeEditor = this.props.editor.get('nativeEditor');
 			const buttonCfg = nativeEditor.config.buttonCfg || {};
 
 			if (Lang.isFunction(buttons)) {
@@ -555,7 +555,7 @@ export default WrappedComponent =>
 			).map(function(button, index) {
 				let props = this.mergeExclusiveProps(
 					{
-						editor: this.context.editor,
+						editor: this.props.editor,
 						key:
 							button.key !== 'separator'
 								? button.key
--- a/src/components/base/button-action-style.js
+++ b/src/components/base/button-action-style.js
@@ -33,7 +33,7 @@ export default WrappedComponent =>
 				Lang.isFunction(this.isActive) &&
 				Lang.isFunction(this.getStyle)
 			) {
-				const editor = this.context.editor.get('nativeEditor');
+				const editor = this.props.editor.get('nativeEditor');
 
 				editor.getSelection().lock();
 
--- a/src/components/base/button-command-active.js
+++ b/src/components/base/button-command-active.js
@@ -24,7 +24,7 @@ export default WrappedComponent =>
 		 * @return {Boolean} True if the command is active, false otherwise.
 		 */
 		isActive() {
-			const editor = this.context.editor.get('nativeEditor');
+			const editor = this.props.editor.get('nativeEditor');
 
 			const command = editor.getCommand(this.props.command);
 
--- a/src/components/base/button-command.js
+++ b/src/components/base/button-command.js
@@ -23,7 +23,7 @@ export default WrappedComponent =>
 		 * @method execCommand
 		 */
 		execCommand = data => {
-			const editor = this.context.editor.get('nativeEditor');
+			const editor = this.props.editor.get('nativeEditor');
 
 			editor.execCommand(this.props.command, data);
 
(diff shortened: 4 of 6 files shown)


src/components/base/toolbar-buttons.js
```diff
@@ -520,7 +520,7 @@ export default WrappedComponent =>
 		getToolbarButtons(buttons, additionalProps) {
 			const buttonProps = {};
 
-			const nativeEditor = this.context.editor.get('nativeEditor');
+			const nativeEditor = this.props.editor.get('nativeEditor');
```
julien, line 523: I'm looking at this again, and I'd really like to understand why `this.context` is always defined the first time the editor is rendered, but is `undefined` after navigating back and loading the editor again.

The only thing I did managed to see is that our "Custom Button" is always defined - I'm going to keep on digging into this

NemethNorbert: What I saw regarding this, is that our AlloyEdtior components have`this.context` with the editor value. After the SPA navigation, our custom button simply has an empty context. When we are calling methods in our custom button from our _base components_, `this.context` remains empty in that scope.
Based on this, the question is why our custom button has an empty context upon SPA navigation?

matuzalemsteles: Taking a look at it today but without much success or discovery, it really is a very strange case I am not sure if it can be related to something from Senna.js yet but some strange things that I have been testing, I call `EditorContext.Consumer` directly on custom button and the funny thing is that it works very well which makes me intrigued because the component with `contextType` is not working as it should, the strangest thing is that it should also happen in the other components.

I will try to investigate this better tomorrow and try to bring something else about it.

matuzalemsteles: I think I have a hypothesis of what is going on, I think this is not really related to Senna.js but it is part of the consequence of the problem.

I think you had the same problem looking at the PTR that Antonio realized that the source code was not called again in the SPA navigation, @jbalsas ended up creating a function to make this function called during the SPA navigation `AlloyEditor.Buttons[ButtonMarquee.key] = AlloyEditor.ButtonMarquee = ButtonMarquee;`, the assumption that we may have a "reference" problem here when browsing with SPA and the above function is called again, neither does the ButtonMarquee source code is executed again and we can have the reference of the previous `class` that was declared first.

I had done a test adding the `contextType` with the context of the AlloyEditor and it is possible to see the context filled in the first time the class is declared, in the navigation with SPA, as the source code of the ButtonMarquee file is not executed again, I do not I can no longer see the filled Button context. this seemed to me to be a problem related to React because it didn't make much sense but it turns out that as the source code was executed the first time calling `AlloyEditor.Base.EditorContext` is an old reference after browsing with SPA and then it is likely that we will see an empty context and maybe the value of EditorContext is empty, the hypothesis is reinforced when you try to call `AlloyEditor.Base.EditorContext.Consumer` in the rendering phase the context is the most recent filled by AlloyEditor, so it is possible to see the value of filled context. This case happens because `contextType` is static so it is not called again during React renderings later.

Even though this already happens without the `contextType` in the ButtonMarquee class, the problem is that the HOCs (`AlloyEditor.Base.ButtonActionStyle`...) that are used to export the class have the same effect as described above and ends up passing an empty context for the components after navigation with SPA.

As this problem exists, a possible solution is instead of adding the HOCs in the standard export of the component is to add it in the function call, so there will be no "reference" problems.

```
export default function () {
	AlloyEditor.Buttons[
		ButtonMarquee.key
	] = AlloyEditor.ButtonMarquee = AlloyEditor.Base.ButtonActionStyle(
		AlloyEditor.Base.ButtonStateClasses(
			AlloyEditor.Base.ButtonStyle(ButtonMarquee)
		)
	);
}
```

This may be related to the `script` being added with `data-senna-track="permanent"`, perhaps adding as `temporary` will force to run the ButtonMarquee source code and remove these "reference" problems.

As a result, I think that making this PR change may not be valid because the real problem is not that.

cc @julien @NemethNorbert

julien: Hey @matuzalemsteles thanks for the information

I just checked and the solution you proposed, i.e.:

Using 
```js
export default function () {
	AlloyEditor.Buttons[
		ButtonMarquee.key
	] = AlloyEditor.ButtonMarquee = AlloyEditor.Base.ButtonActionStyle(
		AlloyEditor.Base.ButtonStateClasses(
			AlloyEditor.Base.ButtonStyle(ButtonMarquee)
		)
	);
}
```

in `main.js` and removing the HOCs from the exports in `ButtonMarquee.jsx` works fine. I think this could be used as a as a workaround - @NemethNorbert could you confirm this is working for you too please? 

If looked at the documentation for the [`ScriptData`](https://github.com/liferay/liferay-portal/blob/f1a2b2ce06fa050d5e1ab193728ab9ec06df4315/portal-kernel/src/com/liferay/portal/kernel/servlet/taglib/aui/ScriptData.java) tag but I couldn't find anything that let's you set the `data-senna-track` attribute on the html `<script>` tag it generates (and it makes sense since `data-senna-track` is only a Senna.js thing).

julien: Another thing I wanted to mention, is that for some scripts, we're [already](https://github.com/liferay/liferay-portal/blob/f1a2b2ce06fa050d5e1ab193728ab9ec06df4315/modules/apps/frontend-editor/frontend-editor-alloyeditor-web/src/main/resources/META-INF/resources/js/alloyeditor.js#L336-L347) changing the `data-senna-track` attribute, these "editor paths" get setup [here](https://github.com/liferay/liferay-portal/blob/f1a2b2ce06fa050d5e1ab193728ab9ec06df4315/modules/apps/frontend-editor/frontend-editor-alloyeditor-web/src/main/resources/META-INF/resources/alloyeditor.jsp#L233-L236), so I don't know if we could re-use that and have a "register" mechanism for custom buttons - just an idea

NemethNorbert: Hey @julien , @matuzalemsteles ,

I tested the suggested workaround and it works just fine. I'll go ahead and share these findings with the customers.

Also at the start of the investigation I modified ScriptData so it printed the script tag requiring the custom button with the "temporary" data-senna-track attribute, however it did not solve the issue for me. I only tested that with the original PoC that did not have a default export in main.js, maybe the combination of the two would work as well.

Anyhow thank you all for working on this issue and finding a solution.

julien:
Closing this PR as we don't need this change. Thanks everyone.

PR #1469, opened 2021-03-08
chore: update to latest version of @liferay/eslint-config

wincent:
Because, as noted in https://github.com/liferay/liferay-frontend-projects/pull/429 we will be turning off the `no-for-of-loops` rule by default, which means we have to turn it back on in places that might still want it, like this repo.

However, while doing this, I figured I might as well update to a newer version of `eslint-config-liferay` in this repo (really, that means moving to `@liferay/eslint-config`). Turns out that this, and updating ESLint like I am here means that we are going to need [this PR](https://github.com/liferay/liferay-frontend-projects/pull/430) too, because the way we monkey-patch ESLint needs to be updated.

Note the `typescript` dependency is added just because some of our rules depend on it, and we happen to use it everywhere else anyway (and back in the day, I didn't want to add a hard-dependency on it to our ESLint config... this is something for us to reevaluate though, probably).

Separate commits apply autofixes for new rules. And another turns on the `no-for-of-loops` rule which is the entire reason I am doing this.

Initially marking this as a draft because it requires two PRs to be merged for `@liferay/eslint-config`, then a release to be cut and rolled into this.


wincent:
Lame. Failing because ESLint and Prettier disagree about where comments go. Going to try and remember how we have resolved that kind of issue in the past.

Three options I'm thinking of:

- Delete the darn comments (in 6 files).
- Turn off the darn lint rule.
- Resolve this the same way we do in liferay-frontend-projects (which is to do formatting with our own npm-scripts wrapper).

wincent:
- Green build: https://travis-ci.org/github/liferay/alloy-editor/builds/762357833
- Red build: https://travis-ci.org/github/liferay/alloy-editor/builds/762357835

Obviously flakey, but the one green one means that it can work, so I am going to merge this.

Issue #1500, opened 2021-09-06
LPS-138431 Error Alloyeditor button-paragraph-align.jsx:56 Cannot read properties of undefined (reading 'icon')


georgel-pop-lr:
Description: If you add 'paragraphAlign' button to Alloy Editor Configurator and you add a paragraph text with an image inside, it will fail to open the menu.  Check https://issues.liferay.com/browse/LPS-138431

Steps: 
1. Use 7.2.x bundle and on your portal-ext.properties add:
editor.wysiwyg.default=alloyeditor
editor.wysiwyg.portal-impl.portlet.ddm.text_html.ftl=alloyeditor
2. Go to Content & Data -> Web content -> Select + Button -> add new Basic Web Content 

3. Select source code and add the content inside this file webcontent_sourcecode.txt

4. Publish the web content and deploy the editor.configurator-1.0.0.jar

5. Open again the web content created at step 2 and try to select a part of text 'This is not going to work' and you will see the error

Actual: Fails to open the menu, and you need to refresh page to continue working with the editor

Expected: Editor menu it should not fail when you select a text next to an image

Note: I used 7.2.x branch git hash d70471addc1516fba7cc4c58b999ee7d0dc02e95

PR #1501, opened 2021-09-06
 fix: Prevent activeCommand in button-paragraph-align.jsx to be undefined (#1500)

georgel-pop-lr:
Fixes: #1500

If you add 'paragraphAlign' button to Alloy Editor Configurator and you add a paragraph text with an image inside, it will fail to open the menu. Check https://issues.liferay.com/browse/LPS-138431

### DIFF
--- a/src/components/buttons/button-paragraph-align.jsx
+++ b/src/components/buttons/button-paragraph-align.jsx
@@ -30,12 +30,13 @@ class ButtonParagraphAlign extends React.Component {
 	render() {
 		let buttonCommandsList;
 		let buttonCommandsListId;
+		const getCommands = this._getCommands();
 
 		if (this.props.expanded) {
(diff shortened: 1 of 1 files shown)


diegonvs (approved):
It partially fixes the error but We introduce another issue here:

`justifyleft` is being set as a default. So, when selecting a "justified center" text, it will show the selection as justify left but it will not crash as https://issues.liferay.com/browse/LPS-138431 describes :)

georgel-pop-lr:
> It partially fixes the error but We introduce another issue here:
> 
> `justifyleft` is being set as a default. So, when selecting a "justified center" text, it will show the selection as justify left but it will not crash as https://issues.liferay.com/browse/LPS-138431 describes :)

Hi @diegonvs,

Could we try to check on CKEditor side if we can fix this?

Regards.

PR #1505, opened 2021-10-12
chore: eslint-plugin migration

kresimir-coko:
Part of https://github.com/liferay/liferay-frontend-projects/issues/677.

I've disabled rules that were spitting errors so that `yarn lint` returns the same warnings as it does on `master`.

This is my first time contributing here, so if I overlooked something, let me know

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,7 +12,7 @@ module.exports = {
 		jest: true,
 		node: true,
 	},
-	extends: ['@liferay/eslint-config/react'],
+	extends: ['plugin:@liferay/react'],
 	globals: {
 		AlloyEditor: true,
 		CKEDITOR: true,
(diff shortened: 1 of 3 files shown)


kresimir-coko:
> @markocikos If this the result of something missing on my side, or do we need to update these dependencies?


I'll investigate a bit later and report back with my findings

kresimir-coko:
I don't get the issue with `eslint` missing, according to `package.json` it should be present. Here's the steps I took:
1. `git clean -dfx`
2. `yarn install` - creates a yarn.lock
3. `npx yarn-deduplicate`
4. `yarn lint` - works without issue

Note: I force-pushed once again because either my initial `npx yarn-deduplicate` didn't do anything (even though it did modify the `yarn.lock` file, or I missed something.

Issue #1529, opened 2022-06-09
Border causes a shifted layout when clicking on the plus icon


orsolyaDekany:
**Do you want to request a _feature_ or report a _bug_?**
Bug

**What is the current behavior?**
Border style causes a layout shift when clicking on the plus icon on the first click.
https://issues.liferay.com/browse/LPS-155757

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
![issue](https://user-images.githubusercontent.com/69793575/172911967-ac116169-fb6d-40ba-ac93-9ef9be37903d.gif)

**What is the expected behavior?**
Layout is not shifted on click.

**Which versions of alloy-editor, and which browser / OS are affected by this issue? Did this work in previous versions?**
All AlloyEditor version
Any Browser/OS

markocikos:
Fixed in https://github.com/liferay/alloy-editor/pull/1530

PR #1530, opened 2022-06-10
fix: modify the toolbar add button border to outline to avoid layout shift

orsolyaDekany:
Hi @liferay-frontend Team,

When clicking on the plus icon of the repeatable HTML field, the layout is shifted on the first click and the new field is not added. I think this is caused by the border style of the toolbar add button. I removed the transparent border and modified the toolbar border width a bit, so the style of the toolbar button has remained the same.

https://github.com/liferay/alloy-editor/issues/1529
LPS: https://issues.liferay.com/browse/LPS-155757

**issue**:
issue
style of the add button before the change:
image

**fix**:
fix
after:
borderstyle-after

Could you please review it?
Thank you.

### DIFF
--- a/src/assets/sass/components/ae-toolbar/skin.scss
+++ b/src/assets/sass/components/ae-toolbar/skin.scss
@@ -23,7 +23,7 @@
 
 .ae-toolbar-add {
     background-color: $toolbar-add-bg-color;
-    border-color: $toolbar-add-border-color;
+    outline-color: $toolbar-add-border-color;
     border-radius: $toolbar-add-border-radius;
     color: $toolbar-add-font-color;
 
--- a/src/assets/sass/components/ae-toolbar/structure.scss
+++ b/src/assets/sass/components/ae-toolbar/structure.scss
@@ -44,7 +44,7 @@
 }
 
 .ae-toolbar-add {
-    border: solid $toolbar-add-border-width transparent;
+    outline: solid $toolbar-add-border-width transparent;
     padding: $toolbar-add-padding;
 
     .ae-button {
--- a/src/assets/sass/skin/atlas/variables/structure.scss
+++ b/src/assets/sass/skin/atlas/variables/structure.scss
@@ -53,7 +53,7 @@ $dropdown-listbox-item-max-height: 44px;
 $dropdown-listbox-item-padding: 8px 12px;
 
 /** TOOLBAR **/
-$toolbar-add-border-width: 2px;
+$toolbar-add-border-width: 1px;
 $toolbar-add-height: 44px;
 $toolbar-add-width: 44px;
 


rolandpakai:
Hi @markocikos 
could you please review this PR?
Thanks, Roland

src/assets/sass/components/ae-toolbar/skin.scss
```diff
@@ -23,7 +23,7 @@
 
 .ae-toolbar-add {
     background-color: $toolbar-add-bg-color;
-    border-color: $toolbar-add-border-color;
+    outline: solid $toolbar-add-border-width $toolbar-add-border-color;
```
markocikos, line 26: Looks like CSS is separated here in AE into style and structure files. Interesting, not sure why. Looks like a lot a hassle for little gain. Maybe the idea is to have style-only skins? In any case, it's probably best to follow the existing pattern, meaning outline width should be in structure. It should also be sorted.

markocikos (commented):
@orsolyaDekany A minor tweak inline, PR in general seems pretty safe and simple. This is not reproducible on `master`, right? The closest to an `HTML` field I see is `Rich Text` field, and that is locked to `ClassicEditor`. Does this change show anywhere else on `master`? Just so we can test it is not breaking normal L&F.

orsolyaDekany:
Hi @markocikos ,
the change does not affect Master. The issue is reproducible with repeatable fields that using AlloyEditor as editor. AlloyEditor cannot be set on Master for repeatable fields, and the issue is not reproducible with CKEditor.

I updated the PR and now the style does not touch the skin. Could you please review it again? Thanks.

src/assets/sass/components/ae-toolbar/skin.scss
markocikos: @orsolyaDekany  It was ok to modify the skin. The pattern of code in AE is to separate styles:
- those that affect layout, like `width`, are in `structure.scss`
- those that do no affect layout, like `color`, are in `skin.scss`

Both are styles on the same `.ae-toolbar-add`.

I think it's best we follow the same pattern of code placement. This does not mean placing all styles in `skin.scss`, as in previous solution, or placing all styles in `structure.scss`, as in current solution.

orsolyaDekany: Hi @markocikos , I see. The structure is a bit confusing. I updated the PR, the color is remained in the skis.scss and other styles are in the sutrcture.scss, I think it is following the pattern. Let me know. Thanks.

orsolyaDekany:
Hi @markocikos , I updated the PR again. Thanks.

diegonvs (approved):
@markocikos is currently on PTO, and your fix looks okay to me. So, I'll merge :)

Issue #1534, opened 2022-07-04
defaultLinkTarget is not used in CKEDITOR.plugins.ae_autolink


bettysteger:
**Do you want to request a _feature_ or report a _bug_?**
_bug_

**What is the current behavior?**
If I set defaultLinkTarget to `_blank` and I paste an URL inside the alloyeditor, it does not add the configured default target, in this case `_blank`. 

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

1. Set the following config (defaultLinkTarget)
2. Paste an URL inside the alloyeditor, it does not add the configured default target, in this case `_blank`. 

```
        buttonCfg: {
          linkEdit: {
            defaultLinkTarget: '_blank'
          }
        }
```

Problem is that the `CKEDITOR.plugins.ae_autolink` only adds the following:

```
if (instance._isValidURL(url)) {
  if (instance._isValidEmail(url)) {
    return '<a href="mailto:' + url + '">' + url + '</a>';
  } else {
    return '<a href="' + url + '">' + url + '</a>';
  }
}
```

**What is the expected behavior?**
The link should open in a new tab, if defaultLinkTarget is set to `_blank`. 

**Which versions of alloy-editor, and which browser / OS are affected by this issue? Did this work in previous versions?**
Alloy-editor 2.14.7

PR #1535, opened 2022-07-05
use defaultLinkTarget when creating a link, fixes autolink

bettysteger:
fixes https://github.com/liferay/alloy-editor/issues/1534

### DIFF
--- a/src/core/link.js
+++ b/src/core/link.js
@@ -90,6 +90,14 @@ Link.prototype = {
 			attrs
 		);
 
+		if (!linkAttrs.target) {
+			const linkCfg = this._editor.config.linkCfg || {};
+			const defaultTarget = linkCfg && linkCfg.defaultTarget;
+			if (defaultTarget) {
+				linkAttrs.target = defaultTarget;
+			}
+		}
+
 		const style = new CKEDITOR.style({
 			attributes: linkAttrs,
 			element: 'a',
--- a/test/core/test/link.js
+++ b/test/core/test/link.js
@@ -606,5 +606,29 @@ describe('Link', function() {
 				'<p>update the url of a <a href="test.com" rel="noopener noreferrer" target="_blank">link</a>.</p>'
 			);
 		});
+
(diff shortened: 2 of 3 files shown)


src/core/link.js
```diff
@@ -90,6 +90,14 @@ Link.prototype = {
 			attrs
 		);
 
+		if(!linkAttrs.target) {
+			const buttonCfg = this._editor.config.buttonCfg || {};
+			const defaultLinkTarget = buttonCfg['linkEdit'] && buttonCfg['linkEdit']['defaultLinkTarget'];
+			if(defaultLinkTarget) {
+				linkAttrs.target = defaultLinkTarget;
+			}
+		}
```
markocikos, line 99: `CKEDITOR.Link` is a low-level utility, for creating and handling links. `buttonCfg` is not an API intended to affect links in general, or link utilities. If is an API to configure specific button components. `linkEdit` is the key identifying [ButtonLinkEdit component](https://github.com/liferay/alloy-editor/blob/master/src/components/buttons/button-link-edit.jsx#L60). We should not couple `linkEdit` with links in general, like in this solution.

I could see this maybe as a new configuration option, on the level of `buttonCfg`. Something like
```js
linkCfg: {
    defaultTarget: '_blank'
}
```
Then, we could read it and merge with other attributes, in `link.js`.

bettysteger: @markocikos thanks for the proposal, i changed it to linkCfg.defaultTarget - makes total sense!

markocikos (changes requested):
@bettysteger Looks good!

Before merging, please do two minor things:
1. Run `yarn format` on repo root, to format the code.
2. We started following [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) on this repo. Please squash and split code into two commits:
   - The first one, with core code modifications, should have a message starting with `feat: `.  It is not a fix, as we are adding new API.
   - The second one, with tests, should have a message starting with `test: `.

markocikos:
see https://github.com/liferay/alloy-editor/pull/1536

Issue #1537, opened 2022-08-25
uibridge isn't resetting definitions once the editor is destroyed and recreated


georgel-pop-lr:
**Do you want to request a _feature_ or report a _bug_?**
Bug

**What is the current behavior?**

Once you call destroy and recreate the editor, when trying to change font or fontsize you get the following error:

`
Uncaught TypeError: Cannot read properties of null (reading 'getRanges') at window.CKEDITOR.window.CKEDITOR.dom.CKEDITOR.command.exec (ckeditor.js?browserId=chrome&minifierType=js&languageId=en_US&t=1655312240360:875:311) at window.CKEDITOR.window.CKEDITOR.dom.CKEDITOR.command.exec (ckeditor.js?browserId=chrome&minifierType=js&languageId=en_US&t=1655312240360:207:171) at c.execCommand (ckeditor.js?browserId=chrome&minifierType=js&languageId=en_US&t=1655312240360:276:251) at a.onClick (ckeditor.js?browserId=chrome&minifierType=js&languageId=en_US&t=1655312240360:878:103) at r._onClick (version.js:1:1) at Object.u (version.js:1:1) at h (version.js:1:1) at m (version.js:1:1) at x (version.js:1:1) at R (version.js:1:1)
`

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

Use this POC sample code to reproduce the issue:
```
....

 <div id="editable2" contenteditable="true" placeholder="Write some content here too">
            <p>Bacon ipsum dolor amet salami cow beef chuck pastrami prosciutto. Ground round landjaeger pancetta, cupim capicola turkey pastrami jowl pork shank chuck flank kevin. Sausage pastrami doner short loin burgdoggen spare ribs. Drumstick picanha pastrami ham hock ground round porchetta. Meatball hamburger cow turducken, flank jowl biltong kielbasa t-bone. Chicken tongue doner, picanha salami pork belly filet mignon brisket strip steak landjaeger venison.</p>
        </div>
        
<button type="button"  onclick="createDestroyEditor()">Create/Destroy</button>


        <script>
            window.ALLOYEDITOR_BASEPATH = 'alloy-editor/';

            var createEditor = true;
            var editor1 = AlloyEditor.editable('description');
            var editor2 = AlloyEditor.editable('editable1');
            var editor3 = AlloyEditor.editable('headline');
            var editor4;

            function createDestroyEditor() {
                if (createEditor) {
[...]

PR #1538, opened 2022-08-25
fix: Reset uibridge definitions each time we destroy and create and editor (#1537)

georgel-pop-lr:
fix: Reset uibridge definitions each time we destroy and create and editor (#1537)

We can use https://issues.liferay.com/browse/LPS-156925 to upgrade alloy-editor version in the portal.


diegonvs (approved):
It looks like unit tests are not being executed, but I took your PR and run unit tests, and everything looks okay. Also, testing manually, I can't see anything broken. So, LGTM

Issue #1542, opened 2022-11-07
Image toolbar is wrongly positioned when adding multiple images at the same time


orsolyaDekany:
**Do you want to request a _feature_ or report a _bug_?**
Bug

**What is the current behavior?**
When dropping 2 images in the editor content and selecting the second image to make the toolbar visible, the toolbar is wrongly positioned, because it is visible on the top of the first image instead of the second one.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
1. Create a web content
2. Drag and drop 2 large images at the same time in the content, so they are positioned vertically
3. Double-click on the second image to make the toolbar visible

**What is the expected behavior?**
The image toolbar should be positioned on the top of the selected image

**Which versions of alloy-editor, and which browser / OS are affected by this issue? Did this work in previous versions?**
Reproduced on version 2.11.8
Not reproduced on version 2.11.7

PR #1543, opened 2022-11-08
fix: Calculate the offsetHeight for the toolbar only when the table is used

orsolyaDekany:
Hi Team,

**Issue:** https://github.com/liferay/alloy-editor/issues/1542
LPS: https://issues.liferay.com/browse/LPS-166604

The issue is not reproduceable before [this commit](https://github.com/liferay/alloy-editor/pull/1404/commits/bef3ce28ff1ba4b5788e511133c3c9e78fcb0549).

**Fix:** the offsetHeight is calculated only when the toolbar is positioned for the table
With the fix, the toolbar is positioned again on top of each image that is currently selected and also on top of every cell/row in a table.

Could you please review it?

cc. @markocikos


markocikos:
@orsolyaDekany Is the bug reproducible in other places where we use alloy editor, or is it specific only to web content on 7.3? I couldn't reproduce it on `master` Blogs page.

src/selections/selection-position.js
```diff
@@ -56,7 +56,7 @@ const centerToolbar = function(toolbar, rect) {
 		if (startContainer.$.nodeType !== Node.ELEMENT_NODE) {
 			startContainer = startContainer.getParent();
 		}
-		if (startContainer) {
+		if (startContainer.$.nodeName === "TD" || "TH") {
```
markocikos, line 59: This will always evaluate truthy
```js
if (false || 'TH') console.log('foo') // foo
```

markocikos (changes requested):
@orsolyaDekany see comment inline, and also https://github.com/liferay/alloy-editor/pull/1543#issuecomment-1307289613

orsolyaDekany:
Hi @markocikos ,
The fix is specific to 7.3.x and the issue is reproducible in Blogs as well. The issue is not reproducible using Alloyeditor in Blogs on Master, because of another root cause that was a unique fix to 7.3.x, caused by [this commit](https://github.com/liferay/liferay-portal-ee/commit/b36a0024b368869aa3182b316b3875c68cf7364d) based on my testing. If that change is reverted, the issue is not reproducible in Blogs anymore, however it is still in Web Content. The different behavior might be because of the different editor configurations between the modules, I am not sure.

In the meantime, I also added `"startContainer.$.nodeName === "TH"`, because I realized that header can be set for the columns and rows after creating a table.

src/selections/selection-position.js
```diff
@@ -56,7 +56,7 @@ const centerToolbar = function(toolbar, rect) {
 		if (startContainer.$.nodeType !== Node.ELEMENT_NODE) {
 			startContainer = startContainer.getParent();
 		}
-		if (startContainer) {
+		if (startContainer.$.nodeName === "TD" || startContainer.$.nodeName === "TH") {
```
markocikos, line 59: It looks like this actually fixes a small bug in `master`, where a text line (paragraph) followed by an image causes wrong positioning. Awesome! I'll trust you that this also fixes the 7.3 bug from the ticket.

Before merging, we should just tidy this a bit. Please move node name to a variable, use single quotes, break line if needed. If you followed [setup](https://github.com/liferay/alloy-editor/wiki/Contributing#setup), you can use `yarn format` (and `yarn test` for doublecheck).

markocikos (changes requested):
@orsolyaDekany almost good to go! Please see comment inline.

markocikos (approved):
LGTM!

